PR #9, opened 2019-12-11
Invitations will automatically be deleted

raresica1234:


src/modules/ModerationModule.cpp
```diff
@@ -0,0 +1,41 @@
...
+		m_invitationModeration ^= true;
+		client.createMessage(message.channelId(), m_invitationModeration ? "Invitations will be deleted!" : "Invitations won't be deleted!");
+    });
+}
+
+void ModerationModule::OnMessage(Discord::Client& client, const Discord::Message& message)
+{
+	Module::OnMessage(client, message);
+	if (m_invitationModeration)
+		if (message.content().startsWith("https://discord.gg"))
```
Gaztin: `startsWith` will fail if someone sends:
> join my server: https://discord.gg/whatever

Gaztin (changes requested):

raresica1234:
That should do it.

PR #11, opened 2020-06-04
Added currency module

NotJustSid:


Gaztin (changes requested):
Please revert all the formatting changes before we continue the review 🙂

src/UmikoBot.h
```diff
@@ -44,16 +52,15 @@ struct CommandInfo {
 	bool adminPermission;
 };
 
-class UmikoBot : public Discord::Client
-{
+class UmikoBot : public Discord::Client {
```
Gaztin: Only revert this part

Gaztin (changes requested):
Reformat with braces after newlines

src/modules/CurrencyModule.cpp
```diff
@@ -1,46 +1,350 @@
...
+	m_timer.setInterval(84600 * 1000); //!24hrs = 84600 seconds
+	QObject::connect(&m_timer, &QTimer::timeout, [this, client]() {
+		
+		//!Clear the daily bonus for everyone
+		for (auto it = m_settings.begin(); it != m_settings.end(); ++it) {
+			it->isDailyClaimed = false;
+		}
+
+		if (!isRandomGiveAwayDone) {
+			client->createMessage(giveawayChannelId, "@everyone Today's freebie expires in **10 seconds**. `!claim` it now!");
```
Gaztin: I don't think we ever want to tag @everyone or even @here with this bot. And I think 10 seconds is a little low. Especially if we don't tag people. Can we make it 30 or 60 instead and skip the tagging?

src/modules/CurrencyModule.cpp
```diff
@@ -1,46 +1,350 @@
...
+			QString time = QString::number(conv.hour()) + "(hrs) " + QString::number(conv.minute()) + "(mins) " + QString::number(conv.second()) + "(secs)";
+			QString desc = "**You have already claimed your daily credits.**\nCome back after `" + time + "` to get more rich!";
+
+			client.createMessage(message.channelId(), desc);
+		}
+		else {
+			m_settings[setting->id].isDailyClaimed = true;
+			m_settings[setting->id].currency += 100;
+
+			client.createMessage(message.channelId(), "**You now have 100 more " + currencyName + " in your wallet!**");
```
Gaztin: Can we make the daily amount variable, and changeable with a command, pretty please? So we don't have to change the source code if we want to balance it in the future.

src/modules/CurrencyModule.cpp
```diff
@@ -1,46 +1,350 @@
...
-	result += "$ " + QString::number(m_settings[user].currency) + "\n";
-	result += "\n";
+void CurrencyModule::OnMessage(Discord::Client& client, const Discord::Message& message) {
+
+	if (!isRandomGiveAwayDone) {
+		static int pingCount = 0;
+		if (pingCount == 0) {
+			bool outcome = randp(random_engine);
+			if (outcome) {
+				client.createMessage(giveawayChannelId, "@everyone **FREEBIE** available now! Go `!claim` some juicy coins!");
```
Gaztin: Here as well, let's not ping @everyone.

src/modules/CurrencyModule.cpp
```diff
@@ -1,46 +1,350 @@
...
+				return;
+
+			}
+			if (message.content().toInt() == selfGambleData.randNum) {
+				m_settings[setting->id].currency += 50;
+				client.createMessage(message.channelId(), "**You guessed CORRECTLY!**\n(**50" + currencySymbol  + "** have been added to your wallet!)");
+			}
+			else {
+				m_settings[setting->id].currency -= 10;
+				client.createMessage(message.channelId(), "**Better Luck next time!**\n*(psst! I took **10CC** from your wallet for my time...)*");
```
Gaztin: Here it says "10CC" but in other places it says "50"+currencySymbol. I assume that's what you want to do here too.

src/modules/CurrencyModule.cpp
```diff
@@ -1,46 +1,350 @@
...
+
+			GuildSetting* setting = &GuildSettings::GetGuildSetting(selfGambleData.guildId);
+			if (message.content().toInt() > 5 || message.content().toInt() < 0) {
+
+				client.createMessage(message.channelId(), "**Your guess is out of range!** \nTry a number between 0 and 5 (inclusive)");
+				return;
+
+			}
+			if (message.content().toInt() == selfGambleData.randNum) {
+				m_settings[setting->id].currency += 50;
```
Gaztin: I think I would like all of these magic numbers as configurable variables. And by "all" I mean minimum roll (0), maximum roll (5), winner reward (50) and loser loss (10).

And since the person is gambling, can't we make it so that they gamble a given amount of currency for a chance to double it? I think that could be exciting!

src/modules/CurrencyModule.h
```diff
@@ -1,23 +1,43 @@
...
+	std::bernoulli_distribution randp{ randGiveawayProb };
+	QTimer m_timer;
+	bool isRandomGiveAwayDone{ false };
+	bool allowGiveaway{ false };
+	snowflake_t giveawayChannelId{0};
+	QString currencyName;
+	QString currencySymbol;
 
-	void StatusCommand(QString& result, snowflake_t guild, snowflake_t user) override;
+	struct selfGamble {
```
Gaztin: I'd appreciate if this was renamed to `SelfGamble` so it isn't confused with being a variable.

Gaztin (changes requested):

src/modules/CurrencyModule.cpp
NotJustSid: Oh I might've missed that, thanks!

src/modules/CurrencyModule.cpp
NotJustSid: Sure, I will get rid of all the magic numbers... Do you want me to make them changeable using commands as well?

NotJustSid: And yeah the doubling thing would be quite exciting, I will work on it now!

src/modules/CurrencyModule.h
NotJustSid: Gotcha! I usually use that as well; I might have typed wrong or something

src/modules/CurrencyModule.cpp
NotJustSid: sure

src/modules/CurrencyModule.cpp
NotJustSid: Aight

NotJustSid:
Found some bugs, will fix them as well...

src/modules/CurrencyModule.cpp
Gaztin: Yes if it is not too much to ask, please make them settable via a command!

src/modules/CurrencyModule.cpp
```diff
@@ -1,46 +1,689 @@
...
+				}
+				else 
+				{
+					client.createMessage(message.channelId(), "**BRUH**, ***yOu CaN't JuSt GeT fReE sTuFf aNyTiMe.***");
+				}
+			}
+			else 
+			{
+				Discord::Embed embed;
+				embed.setColor(qrand()%11777216);
```
Gaztin, line 253: As per the documentation, I would recommend using `RAND_MAX` here since it may very well differ between platforms.

src/UmikoBot.cpp
```diff
@@ -811,4 +719,4 @@ void UmikoBot::GetGuildMemberInformation(snowflake_t guild, snowflake_t after)
 		listGuildMembers(guild, 1000).then(processMembers);
 	else
 		listGuildMembers(guild, 1000, after).then(processMembers);
-}
+}
```
Gaztin: It's unfortunately difficult to see what's actually changed, and what's just been reformatted. I would love to see a diff that's actually representative of what's functionally different.

Gaztin (changes requested):

src/UmikoBot.cpp
Gaztin: (In case it's not clear, I'm talking about this whole file UmikoBot.cpp)

NotJustSid: Changes reverted!

Gaztin (approved):

PR #17, opened 2020-08-14
Implemented a streak system for !daily.

FancyKillerPanda:
It's currently set to increase at 5% of the base level per day (1.00x on day 1, 1.05x on day 2, 1.10x on day 3). If that's too much / too little, I can change it.


Gaztin:
LGTM

PR #21, opened 2020-09-24
Fix save errors for currency.json

NotJustSid:
- `Module::Save(...)` now also truncates the file when opening it.
- Moved some user removal (for users who left the server) code from `CurrencyModule::OnSave(...)`

### DIFF
--- a/src/modules/CurrencyModule.cpp
+++ b/src/modules/CurrencyModule.cpp
@@ -18,28 +18,35 @@
 //! Gamble Timeout in seconds
 #define gambleTimeout 20
 
-CurrencyModule::CurrencyModule(UmikoBot* client)
-	: Module("currency", true), m_client(client)
+CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_client(client)
 {
 
 	m_timer.setInterval(24*60*60*1000); //!24hr timer
 	QObject::connect(&m_timer, &QTimer::timeout, [this, client]() 
 		{
-			for (auto server : guildList.keys()) 
+		for (auto server : guildList.keys()) 
+		{
+			//!Clear the daily bonus for everyone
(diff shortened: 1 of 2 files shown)


src/modules/CurrencyModule.cpp
```diff
@@ -18,28 +18,35 @@
...
-					user->isDailyClaimed = false;
+					user.isDailyClaimed = false;
 				}
-
+				else 
+				{
+					//remove if the user is no longer in the server
+					guildList[server].removeAt(i);
+				}
+			}
```
Gaztin: What happened to the indentation here?

Gaztin (changes requested):

src/modules/CurrencyModule.cpp
```diff
@@ -985,11 +991,8 @@ void CurrencyModule::OnSave(QJsonDocument& doc) const
 	{
 		QJsonObject serverJSON;
 		
-		for (auto user = guildList[server].begin(); user != guildList[server].end(); user++) {
-			if (m_client->GetName(server, user->userId) == "")
-			{
-				continue;
-			}
```
Gaztin, line 992: Why do I get a feeling that this was the bad guy all along..

src/modules/CurrencyModule.cpp
NotJustSid: The indentation is all fine?

NotJustSid: Ok it's fine in the code highlighted here, but it's not in the section that follows (EDIT: and the one just before)... On it.

src/modules/CurrencyModule.cpp
NotJustSid: It's what *does* the bad job, but it isn't what causes the issue...

Gaztin (approved):

PR #22, opened 2020-09-25
Actually fixed the save error

NotJustSid:
Removed the Save() call from ~Module() which truncated the whole file and wrote nothing.

### DIFF
--- a/src/core/Module.cpp
+++ b/src/core/Module.cpp
@@ -11,7 +11,7 @@ Module::Module(const QString& name, bool enabledByDefault)
 
 Module::~Module()
 {
-	Save();
+	
 }
 
 void Module::OnMessage(Discord::Client& client, const Discord::Message& message)


src/modules/CurrencyModule.cpp
```diff
@@ -66,6 +62,17 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+					serverConfig.freebieTimer = new QTimer;
+					serverConfig.freebieTimer->setInterval(serverConfig.freebieExpireTime * 1000);
+					serverConfig.freebieTimer->setSingleShot(true);
+					QObject::connect(serverConfig.freebieTimer, &QTimer::timeout, [this, client, guildId] ()
+						{
+							auto& serverConfig = getServerData(guildId);
+							serverConfig.allowGiveaway = false;
+						});
+					serverConfig.freebieTimer->start();
+				}
```
Gaztin: What happened here?

PR #24, opened 2020-09-26
General Feature Update

NotJustSid:
- `gamble` now displays the value the bot *thought* of when the user loses.
- `gamble` now uses a uniform distribution to *think* of numbers.
- Added `FunUtilModule` to be used for utility and fun commands.
    - Added `meme`


src/UmikoBot.h
```diff
@@ -40,7 +40,9 @@ namespace Commands {
 		CURRENCY_SET_PRIZE_PROB,
 		CURRENCY_SET_PRIZE_EXPIRY,
 		CURRENCY_RICH_LIST,
-		CURRENCY_DONATE
+		CURRENCY_DONATE,
+
+		UFUN_MEME
```
Gaztin, line 45: Why is it called UFUN ?

src/modules/FunUtil.h
```diff
@@ -0,0 +1,12 @@
...
+#include "core/Module.h"
+#include "UmikoBot.h"
+
+class FunUtilModule : public Module, public QObject 
+{
+private:
+	
+public:
+	FunUtilModule();
+};
```
Gaztin, line 12: File should follow the file name convention of the other modules. So in this case the file should be called `FunUtilModule.h`.
That said, I would love to see a little more creative name choice for the module."Util" is usually used when typing up helper functions for an existing subsystem. Since this is going to be its own contained class, I propose something like "FunModule", "FunAndGamesModule", or maybe "EntertainmentModule"!

Gaztin (changes requested):

src/modules/FunUtil.h
NotJustSid: Okay gotcha, but I wanted it to reflect that utility commands can be added to this module.

src/UmikoBot.h
NotJustSid: Utility and Fun...

PR #25, opened 2020-09-27
General Feature Update (with some fixes)

NotJustSid:
- `gamble` now uses a uniform random distribution.
- `gamble` now shows the number the bot *thought* of if the player loses.
- Added a `FunModule`
    - Added `meme` which shows memes from reddit.
- Fixed `isDailyClaimed` was actually not set to `false` at the end of the day.
- Added `compensate` which can be used to pay everyone some amount as a compensation for the downtime during bot maintenance and updates.
- Fixed base `daily` amount starting at `100` instead of the daily set by the server.
- Moved and added information related to the currency (streak and whether the user claimed the daily) to `wallet`.
- Updated `wallet` to take in another optional argument which can be used to view the wallet of other users.


src/modules/FunModule.cpp
```diff
@@ -0,0 +1,61 @@
...
+		if (args.first() != prefix + "meme")
+			return;
+
+		if (args.size() >= 2) {
+			client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
+			return;
+		}
+
+		auto guild = channel.id();
+		QNetworkAccessManager* manager = new QNetworkAccessManager();
```
Gaztin: This is a memory leak. What you can do is store the manager in the FunModule class as a member so we don't have to create the entire subsystem each time someone invokes the !meme comand!

Gaztin (changes requested):

src/modules/FunModule.cpp
NotJustSid: Whoops... Thanks for pointing this out!

src/modules/FunModule.cpp
```diff
@@ -3,9 +3,44 @@
...
+
+			QString in = reply->readAll();
+
+			QJsonDocument doc = QJsonDocument::fromJson(in.toUtf8());
+			auto obj = doc.object();
+			bool isNsfw = obj["nsfw"].toBool();
+			if (isNsfw) {
+				m_MemeManager.get(QNetworkRequest(QUrl("https://meme-api.herokuapp.com/gimme")));
+				return;
+			}
```
Gaztin, line 25: Good catch 👍

Gaztin (approved):

PR #26, opened 2020-09-29
!whois and !iam commands

FancyKillerPanda:
Also has some fixes for the streak/steal system (default values on load).


src/modules/UserModule.h
```diff
@@ -0,0 +1,97 @@
...
+			msg += "**Programming Interests: **" + desc.programmingInterests + "\n";
+		
+		if (desc.currentlyWorkingOn != "")
+			msg += "**Currently working on: **" + desc.currentlyWorkingOn + "\n";
+		
+		if (desc.githubLink != "")
+			msg += "**GitHub: **" + desc.githubLink + "\n";
+
+		return msg;
+	}
```
Gaztin: Prefer multiline function implementations in the source file instead of the header!

src/modules/UserModule.cpp
```diff
@@ -0,0 +1,295 @@
...
+						descriptionData.currentUserDescription->programmingInterests = contentToWrite;
+						break;
+					case 4:
+						descriptionData.currentUserDescription->currentlyWorkingOn = contentToWrite;
+						break;
+					case 5:
+						descriptionData.currentUserDescription->githubLink = contentToWrite;
+						break;
+					}
+				}
```
Gaztin: Is there some way you can make it easier to add more questions in the future? So you only have to add one line, or at least only change one place of the code base?

src/modules/UserModule.cpp
```diff
@@ -0,0 +1,295 @@
...
+			msg += UmikoBot::Instance().GetUsername(channel.guildId(), data.userId);
+			msg += " is currently setting their description.**\nTry again later...";
+			client.createMessage(message.channelId(), msg);
+			return;
+		}
+
+		UserDescription& desc = userDescriptions[channel.guildId()][getUserIndex(channel.guildId(), userId)];
+		QString msg = formDescriptionMessage(desc);
+
+		if (msg == "")
```
Gaztin: Prefer `msg.isEmpty()` instead of comparing against string literal.

Gaztin (changes requested):

src/modules/UserModule.cpp
Gaztin: To elaborate: This switch maps to the question array in the header file. Can this connection be changed so maybe it's a list of pairs with questions strings paired with a `QVariant` or lambda?

src/modules/UserModule.h
```diff
@@ -43,55 +43,24 @@ class UserModule : public Module
...
-		return std::distance(guildDescriptions.begin(), std::prev(guildDescriptions.end()));
-	}
-
-	QString formDescriptionMessage(const UserDescription& desc) const
-	{
-		QString msg;
+#define QUESTION(question, field) std::make_pair(question, [](UserDescription& desc, const QString& value)		\
+	{																											\
+		desc.field = value;																						\
+	})
```
Gaztin, line 51: What happened with the backslashes here?

Gaztin (approved):

PR #27, opened 2020-09-30
Added TODO file.

FancyKillerPanda:


Gaztin:
I don't like the idea of a TODO file that is going to pollute the diff in future PRs. If you want a bugtracker or some kind of kanban planning tool, we can open a GitHub Project!

PR #28, opened 2020-10-02
Steal chance now varies with amount that is being stolen.

FancyKillerPanda:


src/modules/CurrencyModule.cpp
```diff
@@ -1039,24 +1042,47 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			QString thiefName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+			QString victimName = UmikoBot::Instance().GetName(channel.guildId(), victimId);
+		
+			std::random_device device;
+			std::mt19937 prng { device() };
+			std::discrete_distribution<> distribution { { 1 - successChance, successChance } };
+			int roll = distribution(prng);
 			
-			if (qrand() % 100 < config.stealSuccessChance)
+			if (roll)
```
Gaztin, line 1058: Can you explain how you don't have to compare against some target chance any more, and just check if the integer is non-zero?

Gaztin (changes requested):

Gaztin (approved):

PR #29, opened 2020-10-18
Added a Random number command

0xsmft:


src/modules/FunModule.cpp
```diff
@@ -62,4 +64,36 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+	RegisterCommand(Commands::FUN_ROLL, "roll", [this](Discord::Client& client, const Discord::Message& message, const Discord::Channel& channel)
+	{
+
+		QStringList args = message.content().split(' ');
+		GuildSetting* setting = &GuildSettings::GetGuildSetting(channel.guildId());
+		QString prefix = setting->prefix;
+
+
+		if (args.first() != prefix + "roll")
+			return;
```
Gaztin: This is already covered in the command handler!

src/modules/FunModule.cpp
```diff
@@ -62,4 +64,36 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+
+		if (args.size() != 3) {
+			client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
+			return;
+		}
+
+		int max = args.at(2).toDouble();
+		int min = args.at(1).toDouble();
```
Gaztin: I would love it if you could just type `!roll 20` and it would roll 1 to 20 for you. Basically, "min" is 1 if args.size() is 2

Gaztin (changes requested):

src/modules/FunModule.cpp
```diff
@@ -62,4 +64,42 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+		GuildSetting* setting = &GuildSettings::GetGuildSetting(channel.guildId());
+		QString prefix = setting->prefix;
+
+		if (args.size() == 2) {
+			int min = 0;
+			std::random_device rand_device;
+			std::mt19937 gen(rand_device());
+			std::uniform_int_distribution<> dist(min);
+			QString text = QString("My Value was : " + QString::number(dist(gen)));
+			client.createMessage(message.channelId(), text);
```
Gaztin: This code is duplicate of that below. can you just set `min` and fall through?

Gaztin (changes requested):

src/modules/FunModule.cpp
```diff
@@ -66,6 +67,44 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
...
+		}
+
+		if (args.size() == 2) {
+			max = 214748647; // set max to 32 bit limit
+		}
+		else
+		{
+			max = args.at(2).toDouble();
+		}
+		min = args.at(1).toDouble();
```
Gaztin: I think you got it backwards. What I mean was for `min` to be 1 if only one number was passed to the command!
Additionally, please put the curly braces on new lines 🙂

Gaztin (changes requested):

src/modules/FunModule.cpp
```diff
@@ -66,6 +67,46 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
...
+
+		if (args.size() == 2)
+		{
+			min = 1;
+		}
+		else
+		{
+			min = args.at(2).toDouble();
+		}
+		min = args.at(1).toDouble();
```
Gaztin: Think you forgot to swap these out. surely `min` is the first argument in the command? 😛

0xsmft: lol yeah

Gaztin (approved):

PR #30, opened 2020-10-21
General Fixes and Improvements

FancyKillerPanda:


src/modules/CurrencyModule.cpp
```diff
@@ -98,36 +100,52 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			{
+				user = UmikoBot::Instance().GetUserFromArg(channel.guildId(), args, 1);
 
-		Discord::Embed embed;
-		embed.setColor(11777216);
-
-		//! Get User and Sever Data
-		auto guild = channel.guildId();
-		auto config = getServerData(guild);
+				if (user == 0)
```
Gaztin, line 115: Is user an integer?

Gaztin:
I don't wanna be petty. The rest looks good so I will merge it!

PR #32, opened 2020-10-25
Feature update (and code refactoring)

NotJustSid:
- Refactored code in Utility.h
    - Made the functions `inline`.
    - Enclosed the whole thing in a namespace.
    - Moved and introduced some emoji and reaction strings inside utility.
- Added a polling system
    - A `poll` command with various sub-commands.
    - A `give-new-poll-access` command to give poll creation access to some role(s).
    - A `take-new-poll-access` command to take poll creation access from some role(s).


src/modules/FunModule.cpp
```diff
@@ -62,4 +72,712 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+		}
+
+		QStringList args = lines.at(0).simplified().split(' ');
+		GuildSetting* setting = &GuildSettings::GetGuildSetting(channel.guildId());
+		QString prefix = setting->prefix;
+
+		
+
+		if (args.first() != prefix + "poll")
+			return;
```
Gaztin: Doesn't the command handler handle this for us?

src/modules/FunModule.cpp
```diff
@@ -62,4 +72,712 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+	});
+
+	RegisterCommand(Commands::FUN_GIVE_NEW_POLL_ACCESS, "give-new-poll-access", [this](Discord::Client& client, const Discord::Message& message, const Discord::Channel& channel) 
+	{
+		QStringList args = message.content().split(' ');
+		GuildSetting* setting = &GuildSettings::GetGuildSetting(channel.guildId());
+		QString prefix = setting->prefix;
+
+
+		if (args.first() != prefix + "give-new-poll-access") return;
```
Gaztin: Same as above

src/modules/FunModule.cpp
```diff
@@ -62,4 +72,712 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+
+		Permissions::ContainsPermission(client, channel.guildId(), message.author().id(), CommandPermission::ADMIN, [this, args, message, channel](bool result) 
+		{
+			GuildSetting* setting = &GuildSettings::GetGuildSetting(channel.guildId());
+			if (!result) 
+			{
+				UmikoBot::Instance().createMessage(message.channelId(), "**You don't have permissions to use this command.**");
+					return;
```
Gaztin: Why is this indented?

src/modules/FunModule.cpp
```diff
@@ -62,4 +72,712 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+	});
+
+	RegisterCommand(Commands::FUN_TAKE_NEW_POLL_ACCESS, "take-new-poll-access", [this](Discord::Client& client, const Discord::Message& message, const Discord::Channel& channel) 
+	{
+		QStringList args = message.content().split(' ');
+		GuildSetting* setting = &GuildSettings::GetGuildSetting(channel.guildId());
+		QString prefix = setting->prefix;
+
+
+		if (args.first() != prefix + "take-new-poll-access") return;
```
Gaztin: Same as above. should already be handled

src/modules/FunModule.cpp
```diff
@@ -62,4 +72,712 @@ FunModule::FunModule() : Module("funutil", true), m_memeChannel(0)
...
+							{
+								std::size_t total = 0;
+								for (auto& it = poll->options.begin(); it != poll->options.end(); ++it) 
+								{
+									total += it->count;
+								}
+								if (total == poll->maxVotes) 
+								{
+									//! Manually invoke timeout for the poll timer
+									poll->timer->QTimer::qt_metacall(QMetaObject::InvokeMetaMethod, 5, {});
```
Gaztin, line 619: lol

src/modules/FunModule.h
```diff
@@ -2,12 +2,193 @@
...
+							if (serverPolls->at(i)->pollNum == poll_num)
+							{
+								serverPolls->removeAt(i);
+							}
+						}
+					});
+				});
+			});
+		}
+	};
```
Gaztin: I'd prefer it if this was located in the .cpp-file. And since it seems to have a lot of logic behind it, probably its own source files as well.

Gaztin (changes requested):

src/modules/FunModule.cpp
NotJustSid: Ah well I'm pretty sure `LevelModule.cpp` did that for some reason, so I thought it would better to conform to that...
(All other commands in `CurrencyModule.cpp` do that as well)

src/modules/FunModule.cpp
NotJustSid: Hmmm no idea... I will fix it.

src/modules/FunModule.cpp
NotJustSid: Hehe

src/modules/FunModule.h
NotJustSid: Sure

Gaztin (approved):

PR #34, opened 2020-10-29
Added "!bribe" command.

ridhwandll:
Users can now bribe an amount, to get out of the jail. Can be used once per jail period. The success chance increases the more amount user's bribe. The users can't bribe whatever amount they like, there is a max and minimum bribe amount. The admin can change the max, minimum bribe amount and success chance by giving specific command.


src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+						if (args.size() == 2)
+						{
+							auto& config = getServerData(channel.guildId());
+							config.bribeLeastAmount = args.at(1).toInt();
+							client.createMessage(message.channelId(), "Bribe Least Amount set to **" + QString::number(config.bribeLeastAmount) + "**");
+						}
+
+					});
+			});
+	//End*****************************************************************************************************************************************
```
0xsmft: idt we need the start and end comments?

ridhwandll: That's a bribe code indicator, you can easily see that the bribe code stars where and ends where 😄 
That script is huge in size, so I added that.

0xsmft: yeah but it's ez to tell imo

FancyKillerPanda: I agree with beastboioioi, we don't need the start and end comments

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			if (args.size() != 2)
+			{
+				client.createMessage(message.channelId(), "**Wrong Usage of Command!**");
+				return;
+			}
+
+			int remainingJailTime = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)].jailTimer->remainingTime();
+			bool inJail = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)].jailTimer->isActive();
+			bool bribeUsed = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)].isBribeUsed;
+			auto& authorCurrency = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)];
```
FancyKillerPanda: Why not use `authorCurrency` for the other three variables here (e.g. `authorCurrency.isBribeUsed`)?

You could also use the `authorCurrency` variable further below in code to avoid repeatedly finding the user through `getUserIndex()`.

Also, there is a method `getUserData()` that should be used instead of this verbose call to `operator[]` and `getUserIndex()`.

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+
+			QRegExp re("[+]?\\d*\\.?\\d+");
+			if (!re.exactMatch(args.at(1)))
+			{
+				client.createMessage(message.channelId(), "**You can't bribe invalid amounts**");
```
FancyKillerPanda: I know this message is a copy from other places but I don't really like it much, it sounds a tad weird. Something like `"Your bribe amount must be valid!"` could be better maybe?

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+				client.createMessage(message.channelId(), output);
+				return;
+			}
+
+			if (inJail && amountToBribe < config.bribeLeastAmount)
+			{
+				QString leastBribeAmount = QString::number(config.bribeLeastAmount);
+				QString output = QString(
+					":police_officer: **I don't take such a small amount!** :police_officer:\n"
+					"*(You can always give me `%1` or more %2 )*\n"
```
FancyKillerPanda: I think this should be ``""*(You can always give me `%1 %2` or more)*\n"``.
The symbol should be along with the amount (and get rid of the space before the closing parenthesis).

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+				).arg(leastBribeAmount, config.currencySymbol);
+
+				client.createMessage(message.channelId(), output);
+				return;
+			}
+			if (inJail == false)
```
FancyKillerPanda: `if (!inJail)`

Also, if you checked this before the other `if` statements above then you could get rid of the `inJail &&` part of those conditions.

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+				QString output = QString(
+					":police_officer: **Hey you're not in JAIL!** :police_officer:\n"
+				);
+				client.createMessage(message.channelId(), output);
+				return;
+			}
+
+			if (inJail && guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)].currency() - (amountToBribe) < debtMax)
+			{
+				client.createMessage(message.channelId(), ":police_officer: **I will be loved to be bribed with that amount but are you sure that you can give me that amount?** :police_officer:");
```
FancyKillerPanda: This message feels a tad clunky, maybe something like `"I would love to be bribed with that much money but unfortunately you can't afford it."`

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			double successChance = (static_cast<double>(config.bribeSuccessChance)/ (static_cast<double>(config.bribeMaxAmount) * 100)) * amountToBribe;
+
+			QString authorName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+
+			std::random_device device;
+			std::mt19937 prng{ device() };
+			std::discrete_distribution<> distribution{ { 1 - successChance, successChance } };
+			int roll = distribution(prng);
+
+			if (inJail && roll && bribeUsed == false)
```
FancyKillerPanda: You shouldn't need to do the additional checks here if you've returned earlier if they're `false`.

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			else if(inJail && !roll && bribeUsed == false)
+			{
+				guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)].isBribeUsed = true;
+				int totalTime = remainingJailTime + 3600000;
+				authorCurrency.jailTimer->start(totalTime);
+
+				QString time = utility::StringifyMilliseconds(totalTime);
+
+				QString output = QString(
+					":police_officer: **Your bribes don't affect my loyality!** :police_officer:\n"
```
FancyKillerPanda: Spelling mistake (loyalty).

src/modules/CurrencyModule.h
```diff
@@ -105,6 +107,10 @@ class CurrencyModule : public Module
 		int stealFinePercent { 50 };
 		int stealVictimBonusPercent { 25 };
 		int stealFailedJailTime { 3 };
+
+		int bribeMaxAmount{ 150 };
+		int bribeLeastAmount{ 20 };
+		int bribeSuccessChance{ 69 };
```
FancyKillerPanda, line 113: These should be saved and loaded from the currency config file.

FancyKillerPanda (changes requested):

src/modules/CurrencyModule.cpp
FahimFuad: Will I remove it?, if yes how?

src/modules/CurrencyModule.cpp
ridhwandll: yeah, sure sure!

src/modules/CurrencyModule.cpp
ridhwandll: sure

src/modules/CurrencyModule.cpp
ridhwandll: yeah, fixed!

src/modules/CurrencyModule.cpp
ridhwandll: Cool cool, thanks!

src/modules/CurrencyModule.cpp
ridhwandll: Will I just do... roll?

src/modules/CurrencyModule.cpp
ridhwandll: My bad, thanks!

src/modules/CurrencyModule.h
ridhwandll: Sorry I didn't get your word's.....

src/modules/CurrencyModule.cpp
FancyKillerPanda: Yeah just `if (roll)` should be fine if you’ve already returned if the other conditions weren’t true.

src/modules/CurrencyModule.h
FancyKillerPanda: You should make it so those variables are saved to the JSON file with all the other configuration variables. They’d also be loaded at startup so their values would remain even after the bot shuts down.

src/modules/CurrencyModule.cpp
ridhwandll: Okay I removed! Thanks for the suggestions.

src/modules/CurrencyModule.cpp
```diff
@@ -1670,6 +1673,10 @@ void CurrencyModule::OnLoad(const QJsonDocument& doc)
 			config.stealVictimBonusPercent = serverObj["stealVictimBonusPercent"].toString("25").toUInt();
 			config.stealFailedJailTime = serverObj["stealFailedJailTime"].toString("3").toUInt();
 
+			//config.bribeSuccessChance = serverObj["bribeSuccessChance"].toString("30").toUInt();
+			//config.bribeMaxAmount = serverObj["bribeMaxAmount"].toString().toInt();
+			//config.bribeLeastAmount = serverObj["bribeLeastAmount"].toString().toInt();
+
```
ridhwandll: Do I have to add them here?

ridhwandll:
I am doing some cleanups!

PR #35, opened 2020-10-31
Added "!bribe" command

ridhwandll:
Users can now bribe an amount, to get out of the jail. Can be used once per jail period. The success chance increases the more amount user's bribe. The users can't bribe whatever amount they like, there is a max and minimum bribe amount. The admin can change the max, minimum bribe amount and success chance by giving specific command.


0xsmft:
I would change "<hrs, mm, s>to give me more" to "<hrs, mm, s>, To try again" you may not want to include the "," but yeah.

src/modules/CurrencyModule.h
```diff
@@ -84,27 +86,31 @@ class CurrencyModule : public Module
...
+		int dailyBonusAmount{ 50 };
+		int dailyBonusPeriod{ 3 };
+		int stealSuccessChance{ 30 };
+		int stealFinePercent{ 50 };
+		int stealVictimBonusPercent{ 25 };
+		int stealFailedJailTime{ 3 };
+
+		int bribeMaxAmount{ 150 };
+		int bribeLeastAmount{ 20 };
+		int bribeSuccessChance{ 69 };
```
Gaztin: Please don't format code that you didn't write, so that the diff is minimal!

Gaztin (changes requested):

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+
+				client.createMessage(message.channelId(), output);
+				return;
+			}
+
+			if (inJail && authorCurrency.currency() - (amountToBribe) < debtMax)
+			{
+				client.createMessage(message.channelId(), ":police_officer: **I would love to be bribed with that much money but unfortunately you can't afford it.** :police_officer:");
+				return;
+			}
```
Gaztin: ARen't we guaranteed to be in jail if we reach this code? I also think you should use `else if`s where possible so we don't run into dual cases

src/modules/CurrencyModule.h
ridhwandll: Oops, that Visual Studio feature for auto format on copy and pasting. I am fixing this. Thanks! 😄

src/modules/CurrencyModule.cpp
ridhwandll: Yeah, Okay I'm fixing this on next commit 😄

src/modules/CurrencyModule.h
ridhwandll: Fixed!

src/modules/CurrencyModule.cpp
ridhwandll: Done!

ridhwandll (commented):
Gaz all done!

src/modules/CurrencyModule.cpp
```diff
@@ -902,7 +902,219 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+		RegisterCommand(Commands::CURRENCY_BRIBE, "bribe", [this](Discord::Client& client, const Discord::Message& message, const Discord::Channel& channel)
+		{
+
+			QStringList args = message.content().split(' ');
+			auto& config = getServerData(channel.guildId());
+			snowflake_t authorId = message.author().id();
+
+			int remainingJailTime = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)].jailTimer->remainingTime();
+			bool inJail = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)].jailTimer->isActive();
+			auto& authorCurrency = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)];
```
Gaztin: You can move authorCurrency two lines up and use that instead of fetching it multiple times here

Gaztin (changes requested):

src/modules/CurrencyModule.cpp
ridhwandll: Ahh...sure

Gaztin (approved):

PR #36, opened 2020-10-31
Improvements

FancyKillerPanda:
Reduces code duplication (a lot).

### DIFF
--- a/src/UmikoBot.cpp
+++ b/src/UmikoBot.cpp
@@ -10,6 +10,8 @@
 
 #include "Logger.h"
 
+using namespace Discord;
+
 UmikoBot& UmikoBot::Instance()
 {
 	static UmikoBot bot;
@@ -48,10 +50,10 @@ UmikoBot::UmikoBot(QObject* parent)
 		Save();
 	});
 	
-	connect(&getGatewaySocket(), &Discord::GatewaySocket::disconnected, this, &UmikoBot::OnDisconnected);
+	connect(&getGatewaySocket(), &GatewaySocket::disconnected, this, &UmikoBot::OnDisconnected);
 
 	connect(this, &Client::onMessageCreate,
-		[this](const Discord::Message& message)
+		[this](const Message& message)
 	{
 		Q_FOREACH(Module* module, m_modules)
 		{
@@ -59,7 +61,7 @@ UmikoBot::UmikoBot(QObject* parent)
 		}
 
 		getChannel(message.channelId()).then(
-			[this, message](const Discord::Channel& channel)
+			[this, message](const Channel& channel)
 		{
 			GuildSetting setting = GuildSettings::GetGuildSetting(channel.guildId());
 			if (channel.guildId() != 0 && !message.author().bot()) // DM
@@ -77,7 +79,7 @@ UmikoBot::UmikoBot(QObject* parent)
 	});
 
 	connect(this, &Client::onGuildCreate,
-		[](const Discord::Guild& guild)
+		[](const Guild& guild)
 	{
 		GuildSettings::GetGuildSetting(guild.id()); // only creates if it doesn't exist
 	});
@@ -90,14 +92,14 @@ UmikoBot::UmikoBot(QObject* parent)
 	});
 
 	connect(this, &Client::onGuildMemberUpdate,
-		[this](snowflake_t guild, const QList<snowflake_t>& roles, const Discord::User& user, const QString& nick)
+		[this](snowflake_t guild, const QList<snowflake_t>& roles, const User& user, const QString& nick)
 	{
 		m_guildDatas[guild].userdata[user.id()].username = user.username();
 		m_guildDatas[guild].userdata[user.id()].nickname = nick;
 	});
 
 	connect(this, &Client::onGuildRoleUpdate,
-		[this](snowflake_t guild_id, const Discord::Role& role)
+		[this](snowflake_t guild_id, const Role& role)
 	{
 		for (int i = 0; i < m_guildDatas[guild_id].roles.size(); i++)
 			if (role.id() == m_guildDatas[guild_id].roles[i].id()) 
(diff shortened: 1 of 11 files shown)


src/UmikoBot.cpp
```diff
@@ -289,22 +291,16 @@ UmikoBot::UmikoBot(QObject* parent)
...
-				return;
-			}
+		GuildSetting& s = GuildSettings::GetGuildSetting(channel.guildId());
+		QStringList args = message.content().split(' ');
 
-			QStringList args = message.content().split(' ');
-			if (args.first() != s.prefix + "setprefix")
-				return;
+		if (args.first() != s.prefix + "setprefix")
+			return;
```
Gaztin: Isn't this handled in the onMessage before it calls the corresponding command?

src/UmikoBot.cpp
```diff
@@ -741,6 +709,41 @@ QList<Command> UmikoBot::GetAllCommands()
...
+	::Permissions::ContainsPermission(client, channel.guildId(), message.author().id(), CommandPermission::ADMIN, [args, &client, message, channel, requiredNumberOfArgs, argumentShouldBeANumber, callback](bool result)
+	{
+		if (!result)
+		{
+			client.createMessage(message.channelId(), "**You don't have permissions to use this command.**");
+			return;
+		}
+
+		// TODO(fkp): Maybe we should have a parameter specifying if we want exact/at least $x args.
+		// I just don't want to have too many parameters, it's already quite long
```
Gaztin, line 712: Maybe an idea is to send a regex string that needs to be matched, otherwise it will say "Wrong usage of command" ?

src/modules/CurrencyModule.cpp
```diff
@@ -434,350 +433,145 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
-				return;
-			}
-			
-			if (args.size() > 1) 
-			{
-				client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
-				return;
-			}
+		if (args.first() != prefix + "setannouncechan")
+			return;
```
Gaztin: Same as above. This should be handled already.

src/modules/CurrencyModule.cpp
```diff
@@ -1019,223 +810,120 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
-			}
-
-			if (args.size() == 2)
-			{
-				auto& config = getServerData(channel.guildId());
-				config.stealFinePercent = args.at(1).toInt();
-				client.createMessage(message.channelId(), "Steal Fine Amount set to **" + QString::number(config.stealFinePercent) + "%**");
-			}
+		if (args.first() != prefix + "setstealfine")
+			return;
```
Gaztin: Same as above. Should be handled already.

src/modules/CurrencyModule.cpp
```diff
@@ -1019,223 +810,120 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
-			}
-
-			if (args.size() == 2)
-			{
-				auto& config = getServerData(channel.guildId());
-				config.stealFailedJailTime = args.at(1).toInt();
-				client.createMessage(message.channelId(), "Steal Jail Time set to **" + QString::number(config.stealFailedJailTime) + " hours**");
-			}
+		if (args.first() != prefix + "setstealjailhours")
+			return;
```
Gaztin: Same as above. Should be handled already.

src/modules/CurrencyModule.cpp
```diff
@@ -1019,223 +810,120 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
-			}
-
-			if (args.size() == 2)
-			{
-				auto& config = getServerData(channel.guildId());
-				config.dailyBonusAmount = args.at(1).toInt();
-				client.createMessage(message.channelId(), "Daily Bonus Amount set to **" + QString::number(config.dailyBonusAmount) + "**");
-			}
+		if (args.first() != prefix + "setdailybonus")
+			return;
```
Gaztin: Same as above. Should be handled already.

src/modules/CurrencyModule.cpp
```diff
@@ -1019,223 +810,120 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
-			}
-
-			if (args.size() == 2)
-			{
-				auto& config = getServerData(channel.guildId());
-				config.dailyBonusPeriod = args.at(1).toInt();
-				client.createMessage(message.channelId(), "Daily Bonus will occur every **" + QString::number(config.dailyBonusPeriod) + " days**");
-			}
+		if (args.first() != prefix + "setdailybonusperiod")
+			return;
```
Gaztin: Same as above. Should be handled already.

src/modules/CurrencyModule.cpp
```diff
@@ -1019,223 +810,120 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
-				client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
-				return;
-			}
-
 			auto& config = getServerData(channel.guildId());
 			auto amt = args.at(1).toDouble();
 
-			for (auto& user : guildList[channel.guildId()]) 
-			{
+			for (auto& user : guildList[channel.guildId()]) {
```
Gaztin: Please stick to next-line braces

Gaztin (changes requested):

src/UmikoBot.cpp
NotJustSid: Ah well I was supposed to remove all of those redundant conditionals, but I forgot to check the files that were added much before...

Gaztin (approved):

PR #40, opened 2020-11-01
Fixed Crashing and General updates

0xsmft:


Gaztin:
The diff is very hard to read since a lot of it is just whitespace changes. I'd love it if you wouldn't reformat code that you didn't add or change 🙂

PR #41, opened 2020-11-01
Added the Event System!

ridhwandll:
- The `!roll` command is now fixed. The Bot doesn't crash if values greater or less than the 32bit limits!
- Added an event system!


res/commands.json
```diff
@@ -270,6 +270,32 @@
 		"usage": "compensate <amount>",
 		"admin": true
 	},
+	"EVENT": {
+		"brief": "Retrieve information about the running event(s).",
+		"usage": "event",
+		"additional": "**Example: ** `!event` will retrieve information about the running event(s)",
```
FancyKillerPanda: This is kinda unnecessary, it is already described by the `brief` and `usage`.

src/modules/CurrencyModule.h
```diff
@@ -4,6 +4,15 @@
...
 #include <random>
 
+
+//Used for the events (Needed in the EventModule.cpp and CurrencyModule.cpp)
+#define highRiskRewardBonus 60
+#define highRiskRewardStealDecrease 8
+
+#define lowRiskRewardPenalty 50
+#define lowRiskRewardLeastAmoutToSteal 60
+#define lowRiskRewardStealIncrease 20
```
FancyKillerPanda: These should probably be set-able through the bot itself.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+#include <QtCore/qregexp.h>
+#include <QtCore/QtMath>
+
+#include <random>
+
+#define eventConfigLoc QString("eventConfig")
+
+EventModule::EventModule(UmikoBot* client) : Module("event", true), m_client(client)
+{
+	RegisterCommand(Commands::EVENT, "event", [this](Discord::Client& client, const Discord::Message& message, const Discord::Channel& channel)
```
FancyKillerPanda: Reuse of `client` could be confusing.

Why do we even need the `UmikoBot* client` parameter? `UmikoBot` was made a singleton so it can just be accessed through `UmikoBot::Instance().whatever`.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+		auto& config = getEventServerData(channel.guildId());
+		CurrencyModule* currencyModule = static_cast<CurrencyModule*>(UmikoBot::Instance().GetModuleByName("currency"));
+		auto& currencyConfig = currencyModule->getServerData(channel.guildId());
+
+		if (args.size() != 1)
+		{
+			client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
+			return;
+		}
+		if (config.isEventRunning == false)
```
FancyKillerPanda: `if (!config.isEventRunning)`

src/modules/CurrencyModule.cpp
```diff
@@ -1198,7 +1202,55 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			client.createMessage(message.channelId(), output);
+			return;
+		}
+
+		if (amountToSteal < lowRiskRewardLeastAmoutToSteal && eventConfig.eventLowRiskLowRewardRunning)
+		{
+			QString num = QString::number(lowRiskRewardLeastAmoutToSteal);
+			QString desc = QString(
+				"**You can't steal that amount beacuse LowRiskLowReward event is running!**\n"
+				"*(You have to steal more than or equal to `%2 %1`)*\n"
```
FancyKillerPanda: Why is there a minimum amount to steal?

src/modules/CurrencyModule.cpp
```diff
@@ -1198,7 +1202,55 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			).arg(config.currencySymbol, num);
+			client.createMessage(message.channelId(), desc);
+			return;
+		}
+
+		if (roll && eventConfig.eventLowRiskLowRewardRunning)
+		{
+			QString num = QString::number(lowRiskRewardPenalty);
+			victimCurrency.setCurrency(victimCurrency.currency() - amountToSteal);
```
FancyKillerPanda: The victim should only lose the amount that the thief steals.

src/modules/CurrencyModule.cpp
```diff
@@ -1208,10 +1260,10 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
 					":man_detective: **Steal success!** :man_detective:\n"
 					"*%1* has discreetly stolen **`%2 %3`** from under *%4's* nose.\n"
 				).arg(thiefName, stealAmount, config.currencySymbol, victimName);
-				
 			client.createMessage(message.channelId(), output);
 		}
-		else
+
+		else if(!roll)
```
FancyKillerPanda: `else if (!roll)` after `else if (roll)`?

src/modules/CurrencyModule.cpp
```diff
@@ -1239,22 +1291,26 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
 				client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
 				return;
 			}
-
+			if (eventConfig.isEventRunning)
+			{
+				client.createMessage(message.channelId(), "**You cannot set configarations regarding steal while an event is running!** ");
+				return;
+			}
 			if (args.size() == 2)
```
FancyKillerPanda: You've already checked for this.

src/modules/CurrencyModule.cpp
```diff
@@ -1270,22 +1326,26 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
```
FancyKillerPanda: You've already checked for this.

src/modules/CurrencyModule.cpp
```diff
@@ -1301,22 +1361,26 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
```
FancyKillerPanda: You've already checked for this.

src/modules/CurrencyModule.cpp
```diff
@@ -1332,32 +1396,35 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
```
FancyKillerPanda: You've already checked for this.

src/modules/EventModule.h
```diff
@@ -0,0 +1,33 @@
...
+	void OnSave(QJsonDocument& doc) const override;
+	void OnLoad(const QJsonDocument& doc) override;
+	EventConfig& getEventServerData(snowflake_t guild)
+	{
+		return serverEventConfig[guild];
+	}
+private:
+	QMap<snowflake_t, QList<snowflake_t>> m_EventWhitelist;
+	QMap<snowflake_t, EventConfig>serverEventConfig;
+	UmikoBot* m_client;
```
FancyKillerPanda: Why do we need this when `UmikoBot` is a singleton?

src/modules/FunModule.cpp
```diff
@@ -70,31 +70,36 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
...
 		{
 			client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
 			return;
 		}
 
+		double min = args.at(1).toDouble();
+		double max = args.at(2).toDouble();
+
 		QRegExp re("[+-]?\\d*\\.?\\d+");
 		if (!re.exactMatch(args.at(1)) || !re.exactMatch(args.at(2)))
```
FancyKillerPanda: Should probably check if they are doubles before you try converting them.

src/modules/FunModule.cpp
```diff
@@ -70,31 +70,36 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
...
 		{
 			client.createMessage(message.channelId(), "**You must roll with numbers!**");
 			return;
 		}
 
+		//Max and min in 32-bit is 2147483647 and -2147483647
+		if (max >= 2147483647 || min >= 2147483647 || max <= -2147483647 || min <= -2147483647)
```
FancyKillerPanda: Use the constants for `INT_MAX`, etc.

Also why checking `>=` and `<=` instead of `>` and `<`?
Also the negative ones are incorrect, `-2147483648` should be valid.

src/modules/FunModule.cpp
```diff
@@ -70,31 +70,36 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
...
+		{
+			client.createMessage(message.channelId(), "**You can't roll that number!**");
```
FancyKillerPanda: This should be more descriptive, *why* can't they roll that number...

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+						UmikoBot::Instance().createMessage(message.channelId(), "**You don't have permissions to use this command.**");
+						return;
+					}
+					if (args.size() < 2)
+					{
+						UmikoBot::Instance().createMessage(channel.id(), "**Wrong Usage of Command!**");
+						return;
+					}
+					snowflake_t guild = channel.guildId();
+					snowflake_t chan = channel.id();
```
FancyKillerPanda: This is never used.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+					{
+						if (m_EventWhitelist[guild].contains(role)) continue;
+						m_EventWhitelist[guild].push_back(role);
+					}
+
+					UmikoBot::Instance().createMessage(channel.id(), "**The roles have been added.**\nPeople with the role(s) can now launch & end events!");
+
+				});
+		});
+	RegisterCommand(Commands::EVENT_TAKE_NEW_ACCESS, "take-new-event-access", [this](Discord::Client& client, const Discord::Message& message, const Discord::Channel& channel)
```
FancyKillerPanda, line 102: Take *new* event access?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+						return;
+					}
+
+					if (args.size() < 2) {
+						UmikoBot::Instance().createMessage(channel.id(), "**Wrong Usage of Command!**");
+						return;
+					}
+
+					snowflake_t guild = channel.guildId();
+					snowflake_t chan = channel.id();
```
FancyKillerPanda: Never used

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+					QRegExp re("[+]?\\d*\\.?\\d+");
+					if (!re.exactMatch(args.at(2)))
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "**Wrong Usage of Command!**.");
+						return;
+					}
+					if (time > 24)
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "**Currently you can't launch events with an expiry time of more than one day**.");
```
FancyKillerPanda, line 189: Didn't we fix this?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+						embed.setColor(15844367);
+						embed.setTitle("**HighRiskHighReward** event has been launched for `" + num + "` hour(s)!");
+						embed.setDescription("Give command `!event` to see what event is running and its changes!");
+
+						config.isEventRunning = true;
+						config.eventHighRiskHighRewardRunning = true;
+						currencyConfig.stealSuccessChance -= highRiskRewardStealDecrease;
+						UmikoBot::Instance().createMessage(message.channelId(), embed);
+					}
+					if (args.at(1) == "LowRiskLowReward")
```
FancyKillerPanda: What if neither of these are true?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+							}
+							else if (config.eventLowRiskLowRewardRunning)
+							{
+								Discord::Embed embed;
+								embed.setColor(15158332);
+								embed.setTitle("**LowRiskLowReward** event has **ended**!");
+
+								config.isEventRunning = false;
+								config.eventLowRiskLowRewardRunning = false;
+								currencyConfig.stealSuccessChance -= lowRiskRewardStealIncrease;
```
FancyKillerPanda: You shouldn't modify the actual steal success chance variable, as that is saved every few minutes (and it represents the normal steal success chance).

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+				CurrencyModule* currencyModule = static_cast<CurrencyModule*>(UmikoBot::Instance().GetModuleByName("currency"));
+				auto& currencyConfig = currencyModule->getServerData(channel.guildId());
+
+				GuildSetting* setting = &GuildSettings::GetGuildSetting(channel.guildId());
+				if (args.size() != 1)
+				{
+					UmikoBot::Instance().createMessage(message.channelId(), "**Wrong Usage of Command!** ");
+					return;
+				}
+				if (config.isEventRunning == false)
```
FancyKillerPanda: `if (!config.isEventRunning)`

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,463 @@
...
+				{
+					return;
+				}
+
+				if (config.isEventRunning == false)
+				{
+					config.restored = true;
+					return;
+				}
+				if (config.restored == false)
```
FancyKillerPanda: You can simplify a lot of these `if`s.

If you've already checked for one condition, you (usually) don't need to check it again.

FancyKillerPanda (changes requested):

src/modules/EventModule.cpp
ridhwandll: Well you guys use that style, didn't know. I am changing that 😄

src/modules/CurrencyModule.cpp
ridhwandll: Because the LowRiskLowReward event issues a penalty if the user succeeds in stealing! The least amount is there to stop losing money by stealing, suppose you steal 10 CC from a person, you succeeded, now in LowRiskLowReward event you will be charged 50 CC (approximately) for succeeding so you will lose 40 CC if you try to steal 10 CC and succeed! To prevent this the least steal amount is introduced in LowRiskLowReward event! 🙂

src/modules/EventModule.cpp
ridhwandll: Nope, @Sid200 didn't open a PR, he fixed that! 😄

src/modules/CurrencyModule.cpp
ridhwandll: Yes, @FancyKillerPanda , the victim loses that currency! all good here I think.

src/modules/CurrencyModule.cpp
ridhwandll: Should I use only `else` ?

src/modules/CurrencyModule.cpp
ridhwandll: Didn't get you this time 😋

src/modules/EventModule.cpp
ridhwandll: Yes, take event launching and ending access from the people who have the access! (Works like the polls)

src/modules/EventModule.cpp
ridhwandll: Hmm gotta add that, thanks for pointing that out! 😄

src/modules/EventModule.cpp
ridhwandll: I restore the real steal success chance every time the bot starts (after crash)!

src/modules/EventModule.cpp
ridhwandll: Sure!

res/commands.json
ridhwandll: Well...then what should the command `!help event` should display?

src/modules/CurrencyModule.h
ridhwandll: I think that it will be a lot of admin junk commands, like 5 of them here....when more events added...just imagine...

res/commands.json
FancyKillerPanda: Sorry I meant the `additional` bit was unnecessary.

src/modules/EventModule.cpp
FancyKillerPanda: FYI it's not really a matter of style, that's just common C++ practice

src/modules/CurrencyModule.cpp
FancyKillerPanda: Shouldn't the penalty be proportional to the amount stolen?

src/modules/EventModule.cpp
FancyKillerPanda: I still feel like it'd be safer if we just didn't modify those, let's see what @Gaztin says

res/commands.json
ridhwandll: Okay I will remove that 😄

src/modules/EventModule.cpp
ridhwandll: Yes perfectly changed that!

src/modules/CurrencyModule.cpp
ridhwandll: Nice idea, I will try to implement that!

src/modules/EventModule.h
ridhwandll: Yes perfectly removed!

src/modules/EventModule.cpp
ridhwandll: The `guild` is used, but the `chan` is not, I am removing the `chan` 🙂

src/modules/FunModule.cpp
ridhwandll: Will I tell them that it's beyond 32-bit limit? I personally don't like to tell them this 32-bit thingy.

src/modules/CurrencyModule.cpp
ridhwandll: Sorry I didn't get you, I have checked this once in this command.

src/modules/CurrencyModule.cpp
ridhwandll: Sorry I didn't get you, I have checked this once in this command.

src/modules/CurrencyModule.cpp
ridhwandll: Sorry I didn't get you, I have checked this once in this command.

src/modules/FunModule.cpp
ridhwandll: Any need for that? If yes, then how? 😄

src/modules/CurrencyModule.cpp
ridhwandll: Yeah, at last got it. Thanks for pointing that, that was really a big mistake, sorry for that. It is now fixed. 😄

src/modules/FunModule.cpp
ridhwandll: Sure, done the changes!

src/modules/EventModule.cpp
ridhwandll: Can we make a copy of the real `stealSuccessChance` and store that in a variable and tweak the copied variable?

src/modules/CurrencyModule.cpp
ridhwandll: Done!

src/modules/EventModule.cpp
Gaztin: Yeah don't modify those. What you can do is change how we calculate the success chance when a user actually performs a steal command, and use the event system there 😊

ridhwandll: Okay 😄

res/commands.json
```diff
@@ -270,14 +270,40 @@
...
+	},
 	"FUN_MEME": {
 		"brief": "Shows memes from reddit.",
 		"usage": "meme",
 		"admin": false
 	},
 	"FUN_ROLL": {
 		"brief": "Gets a random number.",
-		"usage": "roll <min> <max>",
+		"usage": "roll <min> <max> or !roll <max>",
```
FancyKillerPanda, line 305: I think these should be on separate lines, sort of like how `!help top` does it.

src/modules/CurrencyModule.cpp
```diff
@@ -915,20 +917,123 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+
+		EventModule* eventModule = static_cast<EventModule*>(UmikoBot::Instance().GetModuleByName("event"));
+		auto& eventConfig = eventModule->getEventServerData(channel.guildId());
+
+		if (eventConfig.eventHighRiskHighRewardRunning)
+		{
+			double successChance = (config.highRiskRewardStealSuccessChance / 100.0) * qExp(-0.0001 * qPow(amountToSteal, 1.5));
+
+			QString thiefName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+			QString victimName = UmikoBot::Instance().GetName(channel.guildId(), victimId);
```
FancyKillerPanda, line 934: You already created these above.

src/modules/CurrencyModule.cpp
```diff
@@ -915,20 +917,123 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			}
+			return;
+		}
+
+		else if (eventConfig.eventLowRiskLowRewardRunning)
+		{
+			double successChance = (config.lowRiskRewardStealSuccessChance / 100.0) * qExp(-0.0001 * qPow(amountToSteal, 1.5));
+
+			QString thiefName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+			QString victimName = UmikoBot::Instance().GetName(channel.guildId(), victimId);
```
FancyKillerPanda, line 984: You already created these above.

src/modules/CurrencyModule.cpp
```diff
@@ -915,20 +917,123 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+		QString thiefName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+		QString victimName = UmikoBot::Instance().GetName(channel.guildId(), victimId);
+
+		auto& victimCurrency = guildList[channel.guildId()][getUserIndex(channel.guildId(), victimId)];
+		auto& authorCurrency = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)];
+
+		EventModule* eventModule = static_cast<EventModule*>(UmikoBot::Instance().GetModuleByName("event"));
+		auto& eventConfig = eventModule->getEventServerData(channel.guildId());
+
+		if (eventConfig.eventHighRiskHighRewardRunning)
```
FancyKillerPanda, line 929: (This comment applies to a lot of the steal function, i haven't quoted it all for compactness sake).

I feel like it would be better to calculate the variables once and then have the steal logic execute to avoid code duplication. Instead of copy-pasting the steal code in these `if` statements a bunch of times, we should instead create local variables for the success chance, victim bonus, thief penalty, etc. Then have the steal logic run with those variables, so you don't need to have the steal logic copied many times here.

src/modules/CurrencyModule.cpp
```diff
@@ -915,20 +917,123 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+				double bonus = amountToSteal * highRiskRewardBonus * 0.01;
+				victimCurrency.setCurrency(victimCurrency.currency() - amountToSteal);
+				authorCurrency.setCurrency(authorCurrency.currency() + amountToSteal + bonus);
+				QString num = QString::number(bonus);
+
+				QString stealAmount = QString::number(amountToSteal);
+				QString output = QString(
+					":man_detective: **Steal success!** :man_detective:\n"
+					"*%1* has discreetly stolen **`%2 %3`** from under *%4's* nose.\n"
+					"HighRiskHighReward event **BONUS**: `%5 %3`\n"
```
FancyKillerPanda, line 953: You can just append this part of the message to the `outputString` or whatever after the main portion of the steal logic is done.

src/modules/CurrencyModule.cpp
```diff
@@ -915,20 +917,123 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+
+			QString thiefName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+			QString victimName = UmikoBot::Instance().GetName(channel.guildId(), victimId);
+
+			std::random_device device;
+			std::mt19937 prng{ device() };
+			std::discrete_distribution<> distribution{ { 1 - successChance, successChance } };
+			int roll = distribution(prng);
+			qDebug() << "local success chance"<< successChance;
+			qDebug() << "steal success chance"<< config.stealSuccessChance;
```
FancyKillerPanda, line 941: These shouldn't be here in the final build.

src/modules/CurrencyModule.cpp
```diff
@@ -915,20 +917,123 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+			QString thiefName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+			QString victimName = UmikoBot::Instance().GetName(channel.guildId(), victimId);
+
+			std::random_device device;
+			std::mt19937 prng{ device() };
+			std::discrete_distribution<> distribution{ { 1 - successChance, successChance } };
+			int roll = distribution(prng);
+
+			qDebug() << "local success chance" << successChance;
+			qDebug() << "steal success chance" << config.stealSuccessChance;
```
FancyKillerPanda, line 992: Same as above

FancyKillerPanda (changes requested):
Please make sure you test that everything still works after rebasing.

src/modules/CurrencyModule.cpp
ridhwandll: Yes, I have done this (kind of in the new PR)

src/modules/CurrencyModule.cpp
ridhwandll: Yea, removed these!

src/modules/CurrencyModule.cpp
ridhwandll: Same as above

PR #42, opened 2020-11-08
Added event System!

ridhwandll:
# Added basic Event System
- The events are short time tweaks to some normal configurations or other added functionality to increase the *FUN* 
- Admins can select who can launch or end an event by `!give-new-event-access @Roles`!
- Currently there are two events, `HighRiskHighReward` and `LowRiskLowReward`, they change the copy of the real `stealSuccessChance`!
- Time of an event can be set by the third argument of the `!lanuch` command, events will be automatically ended when the time ends! The `@Roles` who have permission to launch and end events can end the events forcibly by `!endevent`
- The event system has its separate module!
# Update
- New event RaffleDraw!
- The admins and the selected @roles can launch this event by `!launch RaffleDraw <time>`
- The users can buy tickets by `!buyticket <amount>` when the RaffleDraw event starts!
- As always, the admins and the selected @roles can end this event forcibly by !endevent.
- When the event ends, the bot will announce the lucky ticket, and the user have to get the rewards by `!getreward`
- If the user fails to get the reward within 24 hours the reward will be gone, and the admin and the selected @roles can launch RaffleDraw again!
- Currently the price of the RaffleDraw is 500 CC, It can be updated later.
## Fix
- A fix and some improvements for the `!roll` command is added in this pull request. (Finally it won't crash anymore). 
- Some improvement in the bots response in the `!bribe` command.
# How to use this event system?
- To start an event write `!launch <event-code-or-name> <expiry-time-in-hours>`. There are currently 3 events.
- HighRiskHighReward(code to launch: `HRHR`)
   - The steal chance is **DECREASED** for the time being! If the thief succeeds in the steal, he will get a **BONUS** CC which is proportional to the amount he has stolen.
- LowRiskLowReward (code to launch: `LRLR`)
   - The steal chance is **INCREASED** for the time being! If the thief succeeds in the steal, he will have to give a **PENALTY** which is proportional to the amount he has stolen.
- RaffleDraw (code to launch:[No code] just type `RaffleDraw` to launch)
   - The most **EXCITING** event in my opinion!<br/>
   The users can buy tickets with the command `!buyticket <amount of tickets>` (This command will be moved to the shop, when the shop comes!) Each ticket costs `50 CC` initially, this can be changed by the admin command `!setticketprice <amount>`(another thing that will be gone when the shop comes). The users can buy `20` tickets at max initially, can be set buy using the admin command `!setusermaxticket <amount>`. *Each ticket has an unique id*
   - **So why buy this tickets?**
      - The result of the draw will be announced automatically when the event ends, here the result means the ticket id.
      - The user who has the lucky ticket can use the command `!getreward` to get their rewards. Currently the reward is `500 CC` that can be updated later when we get the shop(maybe we can giveaway some shop items?)
   - **NOTE: Admins cannot launch another RaffleDraw event after the end of a RaffleDraw until the lucky user claims their reward. The lucky user have 24 hours to claim their reward, if they don't collect it within 24 hours, rewards will be gone and the admins can start the RaffleDraw event again. *The admins can start this event immidiatley after the lucky user have claimed their reward.* The other events such as the HRHR or LRLR can be launched. They have no relation with the RaffleDraw reward is claimed or not.**
## `!endevent` command
- The event will be ended forcibly! If the RaffleDraw was running then it will announce the lucky ticket immidiatedly. **Can only be used by the `@roles` who have permission to launch or endevents!**
### Event Access management(the roles to launch or end events)
- Use the command `!give-new-event-access @Role` to give access to a certain role(They can launch or end events).
- Use command `!take-new-event-access @Role` to take access from a Role(they can no longer launch or end events).
- **NOTE: These are only the admin commands, only the admins can use this.**


src/modules/CurrencyModule.cpp
```diff
@@ -914,21 +915,113 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+		QString thiefName = UmikoBot::Instance().GetName(channel.guildId(), authorId);
+		QString victimName = UmikoBot::Instance().GetName(channel.guildId(), victimId);
+
+		auto& victimCurrency = guildList[channel.guildId()][getUserIndex(channel.guildId(), victimId)];
+		auto& authorCurrency = guildList[channel.guildId()][getUserIndex(channel.guildId(), authorId)];
+
+		EventModule* eventModule = static_cast<EventModule*>(UmikoBot::Instance().GetModuleByName("event"));
+		auto& eventConfig = eventModule->getEventServerData(channel.guildId());
+		
+		if (eventConfig.eventHighRiskHighRewardRunning)
```
Gaztin: When is this event enabled exactly?

ridhwandll: # HighRiskHighReward
**The steal chance is decreased, but if the users succeed on stealing then they will get BONUS CC**
- The BONUS is proportionate to the amount of the CC that a user steals.
- It enables when you give the command `!launch HRHR <amount-of-time-in-hours>`, Launching all the event is same, you just have to remember the event name 🙂

Gaztin: *When* is it enabled, though?

Gaztin: I can't see where you change this boolean.

Gaztin: Never mind, the diff was closed

Gaztin (changes requested):
I'm sorry, but with the raffle commands, it's a tad too much to review. Please put those commits in another branch revert those commits from this PR.

ridhwandll:
@Gaztin I AM SORRY, because I didn't make them as different, I will follow this from next time!

res/commands.json
```diff
@@ -270,16 +270,69 @@
...
+		"usage": "take-new-event-access <role-mention(s)>",
+		"admin": true
+	},
+	"EVENT_SET_TICKET_PRICE": {
+		"brief": "Sets the ticket price of the RaffleDraw.",
+		"usage": "setticketprice <amount>",
+		"admin": true
+	},
+	"EVENT_SET_USER_MAX_TICKET": {
+		"brief": "Sets the max ticket amount a user can buy.",
```
NotJustSid: I might be being picky, but I'd say change this to
```
Sets the maximum tickets a user can buy.
```
Clean and concise...

src/core/Utility.h
```diff
@@ -81,6 +81,11 @@ namespace utility
 				};
 			}
 			
+			constexpr auto TICKETS = ":tickets:";
+			constexpr auto REGIONAL_INDICATOR_E = ":regional_indicator_e:";
+			constexpr auto REGIONAL_INDICATOR_V = ":regional_indicator_v:";
+			constexpr auto REGIONAL_INDICATOR_N = ":regional_indicator_n:";
+			constexpr auto REGIONAL_INDICATOR_T = ":regional_indicator_t:";
```
NotJustSid: I'd say add all of the `REGIONAL_INDICATORS` instead of just these.. Those are already under `reacts::` so you need to just copy and paste them.

src/modules/CurrencyModule.h
```diff
@@ -4,10 +4,15 @@
 #include "UmikoBot.h"
 #include <random>
 
+//Used for the events (Needed in the EventModule.cpp and CurrencyModule.cpp)
+#define highRiskRewardBonus 50 // In percentage
+#define lowRiskRewardPenalty 40 //In percentage
+
```
NotJustSid: These defines would be better suited under `EventModule.h`

src/modules/CurrencyModule.h
```diff
@@ -110,6 +115,9 @@ class CurrencyModule : public Module
 		int bribeMaxAmount { 150 };
 		int bribeLeastAmount { 20 };
 		int bribeSuccessChance { 68 };
+
+		int lowRiskRewardStealSuccessChance{ 50 };
+		int highRiskRewardStealSuccessChance{ 30 };
```
NotJustSid, line 116: Again, semantically speaking, these variables shouldn't be here.

src/modules/EventModule.h
```diff
@@ -0,0 +1,72 @@
...
+	struct EventConfig
+	{
+		QTimer* eventTimer{ nullptr };
+		QTimer* raffleDrawRewardClaimTimer{ nullptr };
+		bool isEventRunning{ false };
+		bool eventHighRiskHighRewardRunning{ false };
+		bool eventLowRiskLowRewardRunning{ false };
+		bool claimedReward = true;
+		bool eventRaffleDrawRunning{ false };
+		unsigned int currentTicketIndex{ 0 };
```
NotJustSid: This variable stands out. What is this used for?

src/modules/EventModule.h
```diff
@@ -0,0 +1,72 @@
...
+			:m_UserId(userID){}
+	};
+
+	struct EventConfig
+	{
+		QTimer* eventTimer{ nullptr };
+		QTimer* raffleDrawRewardClaimTimer{ nullptr };
+		bool isEventRunning{ false };
+		bool eventHighRiskHighRewardRunning{ false };
+		bool eventLowRiskLowRewardRunning{ false };
```
NotJustSid: You're not required to prefix the variables with "event". A better name would be `isHighRiskHighRewardRunning`. The "is" also clarifies that this is indeed a boolean.

src/modules/EventModule.h
```diff
@@ -0,0 +1,72 @@
...
+		QList<unsigned int> m_TicketIds;
+		RaffleDraw(snowflake_t userID)
+			:m_UserId(userID){}
+	};
+
+	struct EventConfig
+	{
+		QTimer* eventTimer{ nullptr };
+		QTimer* raffleDrawRewardClaimTimer{ nullptr };
+		bool isEventRunning{ false };
```
NotJustSid, line 26: Isn't this redundant given that you have other variables to know if an event is running and what type of event it is?

src/modules/EventModule.h
```diff
@@ -0,0 +1,72 @@
...
+	void OnMessage(Discord::Client& client, const Discord::Message& message) override;
+
+	void OnSave(QJsonDocument& doc) const override;
+	void OnLoad(const QJsonDocument& doc) override;
+private:
+	QMap<snowflake_t, QList<RaffleDraw>> raffleDrawGuildList;
+	QMap<snowflake_t, QList<snowflake_t>> m_EventWhitelist;
+	QMap<snowflake_t, EventConfig>serverEventConfig;
+
+	snowflake_t raffleDrawGetUserIndex(snowflake_t guild, snowflake_t id)
```
NotJustSid: Why does this return a `snowflake_t` (which the code base uses for ids) when it is supposed to return an `int` (the type of the index of elements in `QList`)?

src/modules/FunModule.cpp
```diff
@@ -70,45 +70,74 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
 
 	RegisterCommand(Commands::FUN_ROLL, "roll", [this](Client& client, const Message& message, const Channel& channel)
 		{
+			QStringList args = message.content().split(' ');
```
NotJustSid, line 114: Okay, why are changes to the FunModule which are totally unrelated to the event system in this PR?

src/modules/EventModule.h
```diff
@@ -0,0 +1,72 @@
...
+		QTimer* eventTimer{ nullptr };
+		QTimer* raffleDrawRewardClaimTimer{ nullptr };
+		bool isEventRunning{ false };
+		bool eventHighRiskHighRewardRunning{ false };
+		bool eventLowRiskLowRewardRunning{ false };
+		bool claimedReward = true;
+		bool eventRaffleDrawRunning{ false };
+		unsigned int currentTicketIndex{ 0 };
+		int raffleDrawTicketPrice{ 50 };
+		int maxTicketOfUser{ 20 };
```
NotJustSid: `maxUserTickets` would be a better name...

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+			return;
+		}
+		if (config.eventHighRiskHighRewardRunning)
+		{
+			int num = config.eventTimer->remainingTime();
+			QString time = utility::StringifyMilliseconds(num);
+
+			Embed embed;
+			embed.setColor(15844367);
+			embed.setTitle("**HighRiskHighReward** event");
```
NotJustSid: If I recall correctly, bold has little to no effect in embed titles.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+		}
+		if (config.eventLowRiskLowRewardRunning)
+		{
+			int num = config.eventTimer->remainingTime();
+			QString time = utility::StringifyMilliseconds(num);
+
+			Embed embed;
+			embed.setColor(15844367);
+			embed.setTitle("**LowRiskLowReward** event");
+			embed.setDescription("The steal chance is **increased** but you have to give a **PENALTY** if you succeed!\n"
```
NotJustSid: "pay a penalty" sounds better 😄

src/modules/EventModule.cpp
NotJustSid: I also think "HighRiskHighReward" (all together) in a title would look cluttered.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+
+		}
+		if (config.eventLowRiskLowRewardRunning)
+		{
+			int num = config.eventTimer->remainingTime();
+			QString time = utility::StringifyMilliseconds(num);
+
+			Embed embed;
+			embed.setColor(15844367);
+			embed.setTitle("**LowRiskLowReward** event");
```
NotJustSid: Same as above.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+			return;
+		}
+		if (config.eventRaffleDrawRunning)
+		{
+			int num = config.eventTimer->remainingTime();
+			QString time = utility::StringifyMilliseconds(num);
+
+			Embed embed;
+			embed.setColor(15844367);
+			embed.setTitle("**RAFFLE DRAW** event");
```
NotJustSid: Same as the comments about the titles. Also, let's not capitalize the name.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+			embed.setDescription("You can buy tickets and earn cool rewards if you win!\n"
```
NotJustSid, line 71: We haven't really decided about the rewards have we? The reward is just some cash prize (which I think should be a small amount because we already have freebies).

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+								"The results of the draw is announced when the event expires.\n"
```
NotJustSid: Typo: "results" instead of "result"

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+								"Use `!buyticket <amount>` to buy ticket for participating.\n"
```
NotJustSid: Since `buyticket` let's one buy multiple tickets, use "tickets" instead of "ticket".

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+								"Buying more tickets increases your success chance!\n"
+								"Event **ends** in `" + time + "`");
```
NotJustSid: That's an unnecessary and awkward looking bold in between.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+
+			client.createMessage(message.channelId(), embed);
+		}
+
+	});
+	RegisterCommand(Commands::EVENT_GIVE_NEW_ACCESS, "give-new-event-access", [this](Client& client, const Message& message, const Channel& channel)
+	{
+		QStringList args = message.content().split(' ');
+		UmikoBot::VerifyAndRunAdminCmd(client, message, channel, 2, args, false, [this, &client, channel, message, args]()
```
NotJustSid: You don't use `client` and `args` in the lambda... Why are these in the capture?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+				}
+
+				UmikoBot::Instance().createMessage(channel.id(), "**The roles have been added.**\nPeople with the role(s) can now launch & end events!");
+			});
+	});
+	RegisterCommand(Commands::EVENT_TAKE_NEW_ACCESS, "take-new-event-access", [this](Client& client, const Message& message, const Channel& channel)
+	{
+		QStringList args = message.content().split(' ');
+
+		UmikoBot::VerifyAndRunAdminCmd(client, message, channel, 2, args, false, [this, &client, channel, message, args]()
```
NotJustSid: Same as above.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+
+					if (args.size() != 3)
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "**Wrong Usage of Command!** ");
+						return;
+					}
+
+					if (config.isEventRunning)
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "An event is already running. Please stop that to start a new event!");
```
NotJustSid, line 163: I would still say it would be awesome if we could have multiple events simultaneously.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+
+					QRegExp re("[+]?\\d*\\.?\\d+");
+					if (!re.exactMatch(args.at(2)))
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "**Wrong Usage of Command!**.");
+						return;
+					}
+					if (time > 576)
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "**You can't launch events with an expiry time of more than 24 days!**.");
```
NotJustSid: Shorter message: "Events cannot last longer than 24 days!"

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+
+					double time = args.at(2).toDouble();
+
+					QRegExp re("[+]?\\d*\\.?\\d+");
+					if (!re.exactMatch(args.at(2)))
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "**Wrong Usage of Command!**.");
+						return;
+					}
+					if (time > 576)
```
NotJustSid: Better to write it like `24*24` to make it clear why you have this number here.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+						return;
+					}
+					if (config.eventTimer != nullptr) //Delete the previous timer
+					{
+						delete config.eventTimer;
+						config.eventTimer = nullptr;
+					}
+					config.eventTimer = new QTimer;
+
+					config.eventTimer->setInterval(time * 3600000);
```
NotJustSid, line 187: Again, how does `3600000` come up?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+					config.eventTimer->setSingleShot(true);
+
+					if (args.at(1) != "HRHR" && args.at(1) != "LRLR" && args.at(1) != "RaffleDraw")
```
NotJustSid: To make this extensible and to help conveniently add new events in the future, you should instead make a list of all the codes/names and then check if the argument is in there.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+					{
+						UmikoBot::Instance().createMessage(message.channelId(), "`" + args.at(1) + "` is not an event name or code!");
+						return;
+					}
+					QString eventEmoji = QString(utility::consts::emojis::REGIONAL_INDICATOR_E) + " " +QString(utility::consts::emojis::REGIONAL_INDICATOR_V) + " " +
+													QString(utility::consts::emojis::REGIONAL_INDICATOR_E) + " " + QString(utility::consts::emojis::REGIONAL_INDICATOR_N) + " " +
+															QString(utility::consts::emojis::REGIONAL_INDICATOR_T);
+
+					if (args.at(1) == "HRHR")
```
NotJustSid, line 201: You said we could also use the event names. Where do you handle the case where `args.at(1)` is *"HighRiskHighReward"*?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+						Embed embed;
+						embed.setColor(15844367);
+						embed.setTitle(eventEmoji);
+						embed.setDescription("**HighRiskHighReward** event has been launched for `" + num + "` hour(s)!\nGive command `!event` to see what event is running and its changes!");
+
+						config.isEventRunning = true;
+						config.eventHighRiskHighRewardRunning = true;
+						UmikoBot::Instance().createMessage(message.channelId(), embed);
+					}
+					if (args.at(1) == "LRLR")
```
NotJustSid, line 214: Same query as above.

src/modules/EventModule.cpp
NotJustSid: Also, are you going to implement the idea of creating a wrapper around a list of `QTimer`s to accommodate for larger intervals as we had discussed?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+						Embed embed;
+						embed.setColor(15844367);
+						embed.setTitle(eventEmoji);
+						embed.setDescription("**LowRiskLowReward** event has been launched for `" + num + "` hour(s)!\nUse `!event` to see what event is running and its changes!");
+
+						config.isEventRunning = true;
+						config.eventLowRiskLowRewardRunning = true;
+						UmikoBot::Instance().createMessage(message.channelId(), embed);
+					}
+					if (args.at(1) == "RaffleDraw")
```
NotJustSid, line 227: Why doesn't "RaffleDraw" have a corresponding code?

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+							config.isEventRunning = true;
+							config.eventRaffleDrawRunning = true;
+							UmikoBot::Instance().createMessage(message.channelId(), embed);
+						}
+						else
+						{
+							auto& config = getServerEventData(channel.guildId());
+							int timeLeft = config.raffleDrawRewardClaimTimer->remainingTime();
+							QString time = utility::StringifyMilliseconds(timeLeft);
+							UmikoBot::Instance().createMessage(message.channelId(), "**The winner of the previous RaffleDraw didn't claimed their rewards yet.**\n"
```
NotJustSid: "hasn't claimed" instead of "didn't claimed".

src/modules/EventModule.h
NotJustSid: Edit: Okay I see how you use it, but it could be named better. You could go with something like `numticketsDispensed` or `numticketsBought` to signify that it's the total number of tickets that have been bought.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+								//Set the reward claiming time (24 hours)
+								config.raffleDrawRewardClaimTimer = new QTimer;
+								config.raffleDrawRewardClaimTimer->setInterval(24 * 3600000);
+								config.raffleDrawRewardClaimTimer->setSingleShot(true);
+
+								QObject::connect(config.raffleDrawRewardClaimTimer, &QTimer::timeout, [this, guildID, chan, authorId]()
+									{
+										auto& config = getServerEventData(guildID);
+										config.claimedReward = true;
+										UmikoBot::Instance().createMessage(chan, "The owner of the price in the RaffleDraw didn't get their rewards!\n"
```
NotJustSid: Should be "prize" instead of "price".

NotJustSid: Also use "claim" instead of "get"

NotJustSid: And it would be better if we don't reference the winner as the owner of the prize because they would have to claim it first to own it.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+						UmikoBot::Instance().createMessage(channel.id(), "**You don't have permissions to end events.**");
+						return;
+					}
+					auto& config = getServerEventData(channel.guildId());
+					CurrencyModule* currencyModule = static_cast<CurrencyModule*>(UmikoBot::Instance().GetModuleByName("currency"));
+					auto& currencyConfig = currencyModule->getServerData(channel.guildId());
+
+					QString eventEmoji = QString(utility::consts::emojis::REGIONAL_INDICATOR_E) + " " + QString(utility::consts::emojis::REGIONAL_INDICATOR_V) + " " +
+											QString(utility::consts::emojis::REGIONAL_INDICATOR_E) + " " + QString(utility::consts::emojis::REGIONAL_INDICATOR_N) + " " +
+												QString(utility::consts::emojis::REGIONAL_INDICATOR_T);
```
NotJustSid, line 313: You don't need to hype up the end of an event.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+								config.claimedReward = true;
+								UmikoBot::Instance().createMessage(chan, "The owner of the price in the RaffleDraw didn't get their rewards!\n"
+																		"The role(s) who have permission can start this event again!");
+							});
+							config.raffleDrawRewardClaimTimer->start();
+							config.isEventRunning = false;
+							config.eventRaffleDrawRunning = false;
+							config.eventTimer->stop(); // Forcefully stop the timer.
+							UmikoBot::Instance().createMessage(message.channelId(), embed);
+						}
```
NotJustSid, line 339: There is a lot of code duplication in end-event. You could instead manually trigger the timer timeout and handle ending in the timeout callback.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+		QStringList args = message.content().split(' ');
+		if (args.size() != 1)
+		{
+			client.createMessage(message.channelId(), "**Wrong Usage of Command!** ");
+			return;
+		}
+		auto& config = getServerEventData(channel.guildId());
+		if (config.isEventRunning)
+		{
+			client.createMessage(message.channelId(), "**You can't get free rewards anytime!**");
```
NotJustSid: The rewards technically aren't free considering you have to pay for the tickets.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+					auto& currencyConfig = currencyModule->getServerData(channel.guildId());
+					auto& authorCurrency = currencyModule->guildList[channel.guildId()][currencyModule->getUserIndex(channel.guildId(), message.author().id())];
+					QString authorName = UmikoBot::Instance().GetName(channel.guildId(), message.author().id());
+					config.claimedReward = true;
+					Embed embed;
+					embed.setColor(15844367);
+					authorCurrency.setCurrency(authorCurrency.currency() + 500);
+					embed.setTitle("**Raffle Draw Reward goes to " + authorName + "!**");
+					embed.setDescription("Congratulations **" + authorName + "**!\n"
+										"You just got `500` " + currencyConfig.currencySymbol);
```
NotJustSid: In almost all the other places the symbol is included inside the code block, so would be better if we did that here as well :)

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+			}
+			else if (config.raffleDrawRewardClaimTimer->isActive())
+			{
+				client.createMessage(message.channelId(), "**BRUH, you don't have the lucky ticket!**");
+				return;
+			}
+		}
+		else
+		{
+			client.createMessage(message.channelId(), "**You can't get free rewards anytime!**");
```
NotJustSid: > The rewards technically aren't free considering you have to pay for the tickets.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+		{
+			client.createMessage(message.channelId(), "**You can't buy ticket(s) in invalid amounts**");
+			return;
+		}
+		unsigned int ticket = args.at(1).toUInt();
+		snowflake_t authorID = message.author().id();
+		auto& configRD = raffleDrawGuildList[channel.guildId()][raffleDrawGetUserIndex(channel.guildId(), authorID)];
+		if (configRD.m_TicketIds.size() >= config.maxTicketOfUser)
+		{
+			client.createMessage(message.channelId(), "**You have already bought so many tickets, I can't give you more!**");
```
NotJustSid: I think both of messages for these checks should be the same (the one below is almost similar, but you forgot a space after the period):
```
"**Don't be selfish!**\nLet others also have a go at some tickets. You can only have " + num + " tickets at max!"
```

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+		auto& authorCurrency = currencyModule->guildList[channel.guildId()][currencyModule->getUserIndex(channel.guildId(), authorID)];
+
+		if (authorCurrency.currency() <= 0)
+		{
+			client.createMessage(message.channelId(), "**Haha you poor bruh. Git gud enough to afford tickets.**");
+			return;
+		}
+		if (totalFee > authorCurrency.currency())
+		{
+			client.createMessage(message.channelId(), "**Can you afford that much ticket?**");
```
NotJustSid: Use "Can you afford that many tickets?" instead.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+		for (int i = 0; i != ticket; i++)
+		{
+			config.currentTicketIndex++;
+			configRD.m_TicketIds.push_back(config.currentTicketIndex);
+		}
+
+		authorCurrency.setCurrency(authorCurrency.currency() - totalFee);
+		QString ticketEmoji = QString(utility::consts::emojis::TICKETS);
+		QString ticketAmount = QString::number(configRD.m_TicketIds.size());
+		client.createMessage(message.channelId(), ticketEmoji + " **You own** `" + ticketAmount + "` **ticket(s) now!**\n *(Buy them more to increase your chance)*");
```
NotJustSid: "Buy them more" sounds weird; use "buy more of them" instead.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+		}
+	});
+	RegisterCommand(Commands::EVENT_SET_TICKET_PRICE, "setticketprice", [this](Client& client, const Message& message, const Channel& channel)
+	{
+		QStringList args = message.content().split(' ');
+
+		UmikoBot::VerifyAndRunAdminCmd(client, message, channel, 2, args, true, [this, &client, channel, message, args]()
+			{
+				auto& config = getServerEventData(channel.guildId());
+				config.raffleDrawTicketPrice = args.at(1).toInt();
```
NotJustSid: Most of the currency system currently uses `double`s (though we have plans to change it I guess), so one would assume that this also accepts `doubles`.

src/modules/EventModule.cpp
```diff
@@ -0,0 +1,792 @@
...
+	});
+	RegisterCommand(Commands::EVENT_SET_USER_MAX_TICKET, "setusermaxticket", [this](Client& client, const Message& message, const Channel& channel)
+	{
+		QStringList args = message.content().split(' ');
+
+		UmikoBot::VerifyAndRunAdminCmd(client, message, channel, 2, args, true, [this, &client, channel, message, args]()
+			{
+				auto& config = getServerEventData(channel.guildId());
+				config.maxTicketOfUser = args.at(1).toInt();
+				client.createMessage(message.channelId(), "User max ticket set to **" + QString::number(config.maxTicketOfUser) + "**");
```
NotJustSid: There aren't any global and user level max ticket amounts, so the "User" here isn't necessary.

src/modules/CurrencyModule.cpp
```diff
@@ -914,21 +915,113 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
...
+				QString output = QString(
+					":rotating_light: **You got caught!** :rotating_light:\n"
+					"*%1* has been fined **`%2 %3`** and placed in jail for %4 hours.\n"
+					"*%5* has been granted **`%6 %3`** in insurance."
+				).arg(thiefName, fineAmount, config.currencySymbol, jailTime, victimName, victimBonus);
+
+				client.createMessage(message.channelId(), output);
+			}
+			return;
+		}
```
NotJustSid: Code duplication here. Most of the code for the three modes - "HRHR", "LRLR" and "Normal" is similar and could be combined. Create local variables instead, set them accordingly with the conditionals, and use them for the whole *success chance and friends* stuff.

NotJustSid (changes requested):

res/commands.json
ridhwandll: Perfectly Done! 🙂

src/core/Utility.h
ridhwandll: Okay I added all of them, (a - z). 🙂

src/modules/CurrencyModule.h
ridhwandll: I think they should be in the `CurrencyModule.cpp`. No other events are using them and they only effect the CurrencySystem. I removed the defines to the `EventModule.h` because they are needed by both of the modules. But in this case I don't think its needed to move them. 🙂

src/modules/EventModule.h
ridhwandll: Nope, its needed because it activates whenever an event is running that can be any event. It tracks the event system is in use or not. The event timer is based on it. 🙂

src/modules/FunModule.cpp
ridhwandll: See the PR comment, the roll command is updated 🙂 
image

src/modules/EventModule.cpp
ridhwandll: Then, you want `High Risk High Reward` ?

src/modules/EventModule.cpp
ridhwandll: Okay done 🙂

src/modules/EventModule.cpp
ridhwandll: Same as above 😋

src/modules/EventModule.cpp
ridhwandll: Yes, I am decrementing it to `400`, we need a prize. I will add them when the shop comes 😄

ridhwandll: EDIT: From the above comment of mine: *add them* means adding items from the shop.

ridhwandll: The cash prize will get decreased heavily, when the shop item comes 😄

src/modules/EventModule.cpp
ridhwandll: Thanks for pointing that out 😄

src/modules/EventModule.cpp
ridhwandll: Oops, forgot to remove them after copying 😋 REMOVED 😄

src/modules/EventModule.cpp
ridhwandll: Same as above.

src/modules/EventModule.cpp
ridhwandll: That would be a complete mess in my opinion. People would mix up the events if we run multiple events simultaneously. The events are unique and kind of a surprise. Multiple events will make them very common thing. 🙂

src/modules/EventModule.cpp
ridhwandll: Nope, 24 days is enough for the events in my opinion.

src/modules/EventModule.cpp
ridhwandll: That's the time in milliseconds

src/modules/EventModule.cpp
ridhwandll: Okay perfectly done 😄

src/modules/EventModule.cpp
ridhwandll: I and FKP decided to make the big event names turn into a code. That *HighRiskHighReward* is very tedious to write. 🙂

src/modules/EventModule.cpp
ridhwandll: Because, it is not tedious to write 😄

src/modules/EventModule.cpp
ridhwandll: It needs to attract the eyes, if I remove the current title of the embed, Please give me a better title and a description. Yes, the description is needed I think because, the only title wouldn't attract the users attention, its so small.

src/modules/EventModule.cpp
ridhwandll: Sorry, I didn't get your words.

ridhwandll: EDIT: I understand the code duplication thing, but I didn't understand what you want me to do.

src/modules/CurrencyModule.cpp
ridhwandll: Okay done

Gaztin (approved):

PR #43, opened 2020-11-15
Changed m_Currency to be stored as an integer number of cents.

JacoMalan1:


src/modules/CurrencyModule.h
```diff
@@ -12,20 +12,20 @@ class CurrencyModule : public Module
...
-		double maxCurrency;
+        int maxCurrency;
 		unsigned int dailyStreak;
 		QTimer* jailTimer;
 		unsigned int numberOfDailysClaimed;
 		unsigned int numberOfGiveawaysClaimed;
 
 		UserCurrency(snowflake_t userId, double currency, double maxCurrency, bool isDailyClaimed, unsigned int dailyStreak, unsigned int numberOfDailysClaimed, unsigned int numberOfGiveawaysClaimed)
-			: userId(userId), maxCurrency(maxCurrency), isDailyClaimed(isDailyClaimed), isBribeUsed(false), dailyStreak(dailyStreak), jailTimer(new QTimer()), numberOfDailysClaimed(numberOfDailysClaimed), numberOfGiveawaysClaimed(numberOfGiveawaysClaimed)
+            : userId(userId), maxCurrency(std::floor(maxCurrency * 100)), isDailyClaimed(isDailyClaimed), isBribeUsed(false), dailyStreak(dailyStreak), jailTimer(new QTimer()), numberOfDailysClaimed(numberOfDailysClaimed), numberOfGiveawaysClaimed(numberOfGiveawaysClaimed)
```
FancyKillerPanda, line 28: Shouldn't this be `maxCurrency(std::floor(maxCurrency / 100.0f))`?

Also might be worth using a proper rounding function instead of `std::floor`

src/modules/CurrencyModule.h
```diff
@@ -12,20 +12,20 @@ class CurrencyModule : public Module
...
-		double m_Currency;
+        int m_Currency;
 
 	public:
 		snowflake_t userId;
 		bool isDailyClaimed;
 		bool isBribeUsed;
-		double maxCurrency;
+        int maxCurrency;
 		unsigned int dailyStreak;
```
ridhwandll, line 22: Can you please fix the indentation? 🙂

ridhwandll (changes requested):

src/modules/CurrencyModule.h
```diff
@@ -70,17 +70,17 @@ class CurrencyModule : public Module
 			return *this;
 		}
 
-		double currency() const
+        double currency() const
 		{
-			return m_Currency;
+            return (double)m_Currency / 100.0;
 		}
```
FancyKillerPanda, line 76: I feel like this should also be changed to be an `int`. Basically changing everything relating to currency which is a `double` to use an `int` (number of cents).

src/modules/CurrencyModule.h
```diff
@@ -70,17 +70,17 @@ class CurrencyModule : public Module
...
 
 		void setCurrency(double value)
 		{
-			m_Currency = value;
+            m_Currency = std::floor(value * 100);
```
FancyKillerPanda, line 80: As above, `value / 100.0f`?

src/modules/CurrencyModule.h
```diff
@@ -70,17 +70,17 @@ class CurrencyModule : public Module
...
 
-			if (maxCurrency < currency())
-				maxCurrency = currency();
+            if (maxCurrency < currency())
+                maxCurrency = std::floor(currency() * 100);
```
FancyKillerPanda, line 83: As above

FancyKillerPanda (changes requested):
You (your editor) has changed some of the indentation from tabs to spaces, please change those back.

PR #45, opened 2020-11-15
Changed currency to be stored as an integer number of cents.

JacoMalan1:
Sorry for the messy diff. My editor messed with the tabs and I had to fix it.


JacoMalan1:
Note: the tabs might seem bugged out, but if you take a look at the actual file, they are fine.

PR #47, opened 2020-11-17
Fix crash for multiple polls

NotJustSid:
Yeah that's it. Nothing to see in the description...

EDIT: Well, to be honest the crash occurs after the first poll ends and subsequent polls try to display their results. So it's not really *concurrent*, but yeah this PR fixes that.

### DIFF
--- a/src/modules/PollSettings.cpp
+++ b/src/modules/PollSettings.cpp
@@ -14,11 +14,12 @@ FunModule::PollSettings::PollSettings(const PollOptions& op, long long maxvotes,
 	timer->start();
 
 	QObject::connect(timer.get(), &QTimer::timeout,
-	[poll_num = pollNum, serverPolls = polls]() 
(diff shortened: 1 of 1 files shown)


FancyKillerPanda (approved):
Looks alright, the algorithm to remove the poll and decrement the rest of the numbers worries me slightly, but I think it should work. I'm just wary about if there's ever a situation where two polls have the same number, you could delete both of them but only decrement once.

Looks good otherwise, other than the `this->` inside the lambda which I'm fairly sure is unnecessary, but not deal-breaking.

NotJustSid:
@FancyKillerPanda I don't think there's anyway (without doing it manually) that the bot assigns the same index to two different polls. So that isn't to be worried about.

The `this->` inside the lambda is of course redundant code-wise, but I explicitly included it in there to prevent any confusion.

PR #49, opened 2020-11-23
Added !github

masterneko:


src/modules/FunModule.cpp
```diff
@@ -539,6 +576,20 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
 
 	});
 
+	RegisterCommand(Commands::FUN_GITHUB, "github", [this](Client& client, const Message& message, const Channel& channel) 
+	{
+		QStringList args = message.content().split(' ');
+
```
ridhwandll, line 586: This variable is used no where, you can check for the correct input and make this variable used.
like
`if(args.size() != 1) { /*send message*/ return; }`

ridhwandll (changes requested):

src/modules/FunModule.cpp
```diff
@@ -50,6 +50,43 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
...
 		});
+
+			QObject::connect(&m_GithubManager, &QNetworkAccessManager::finished,
+		this, [this](QNetworkReply* reply) {
+			auto& client = UmikoBot::Instance();
+
+			if (reply->error()) {
+				qDebug() << reply->errorString();
+				client.createMessage(m_GithubChannel, reply->errorString());
+			}
```
Gaztin: Don't we want to exit out if the response contained an error?

Gaztin (changes requested):

PR #52, opened 2020-11-25
Renamed "!stats" to "!records"

ridhwandll:
## Renamed the `!stats` command to the `!records` command
- Fixed a typo in the `res\commands.json` file. (`!getreward`'s | **price** -->**prize**)
EDIT:
- The command typed now have to match exactly with the bots command.
**Example: `!dailygiveme` will not work anymore. The user have to type the correct command.**


res/commands.json
```diff
@@ -340,7 +340,7 @@
 	"FUN_ROLL": {
 		"brief": "Gets a random number.",
 		"usage": "roll <min> <max>\nroll <max>",
-		"additional": "**Example:\n** `!roll 4 10`: Gets a random number between `4` and `10` (inclusive).\n `roll 10`: Gets a random number from `0` to `10` (inclusive).",
+		"additional": "**Example:\n** `!roll 4 10`: Gets a random number between `4` and `10` (inclusive).\n `!roll 10`: Gets a random number from `0` to `10` (inclusive).",
```
Gaztin, line 343: `!` is not necessarily the prefix being used!

res/commands.json
```diff
@@ -89,9 +89,9 @@
 		"additional": "",
 		"admin:": false
 	},
-	"USER_MODULE_STATS": {
+	"USER_MODULE_RECORDS": {
 		"brief": "Shows statistics about a user's time in the server.",
-		"usage": "stats <Person>",
+		"usage": "records <Person>",
```
Gaztin, line 94: Why did you rename it to "records"?

src/modules/EventModule.cpp
```diff
@@ -462,7 +462,7 @@ EventModule::EventModule(UmikoBot* client) : Module("event", true)
 		auto& config = getServerEventData(channel.guildId());
 		if (configRD.m_TicketIds.size() == 0)
 		{
-			client.createMessage(message.channelId(), "**You have no tickets.**\nYou can buy them by using `!buyticket <amount>`");
+			client.createMessage(message.channelId(), "**You have no tickets.**\nYou can buy them by using `!buytickets <amount>`");
```
Gaztin: Same as earlier. `!` may not be the chosen prefix!

Gaztin (changes requested):

res/commands.json
ridhwandll: It is in the TODO, to change that. I makes confusion with the `!status` command. 🙂

res/commands.json
ridhwandll: It is in the commands.json. I have followed the `!top` command 
image

src/modules/EventModule.cpp
ridhwandll: Okay done 🙂

res/commands.json
Gaztin: ah crap. all right then

ridhwandll:
Okay I am closing this. I will open a new PR shortly 🙂

PR #55, opened 2020-11-28
Proofread README and fix typos

NotJustSid:


README.md
```diff
@@ -56,15 +56,15 @@ Thanks to those, all you need to do is run them in succession and you should hav
...
+On Windows it will ask you for the paths of the **x86** and **x64** versions of Qt and OpenSSL. You don't need both of the versions; you can just specify one and only build for that specific platform.
 
 > Make sure you use forward slashes instead of the backslashes.
 
 #### `generate_project_files.sh`
 
 As the name suggests, this script generates the necessary files by using [Premake](https://premake.github.io/).
 
-You can provide an action like you do with premake, but the script is configured such that it generates file for
+You can provide an action like you do with premake, but the script is configured such that it (by default) generates files for
```
JacoMalan1: This sentence is a bit clunky. I think "by default" should be right after ", but" without brackets. I.E.
...but by default the script...

README.md
```diff
@@ -138,21 +138,21 @@ You have successfully forked and cloned the project. If you came from [Contribut
...
 
-Remember, you can always check the *status* of your repo, using
+Remember, you can always check the *status* of your repo using
 
 ```
 git status
 ```
 
-You can also look into more about commands by using
+You can also find more about a command by using
```
JacoMalan1: "find" should be "find out"

JacoMalan1 (changes requested):

README.md
NotJustSid: The script is configured to do something by default; it's not configured by default to do something. But yeah I guess I'll replace it with something shorter and more elegant.

README.md
NotJustSid: Yep. Thanks for pointing that out!

PR #57, opened 2020-12-06
Updated the Moderation module. The "invitations" command now works!

ridhwandll:
Nothing to Explain.

### DIFF
--- a/src/modules/ModerationModule.cpp
+++ b/src/modules/ModerationModule.cpp
@@ -1,6 +1,6 @@
 #include "ModerationModule.h"
 #include "UmikoBot.h"
-#include "Core/Permissions.h"
+#include "core/Permissions.h"
 
 using namespace Discord;
 
@@ -10,8 +10,12 @@ ModerationModule::ModerationModule()
 	RegisterCommand(Commands::MODERATION_INVITATION_TOGGLE, "invitations",
 		[this](Client& client, const Message& message, const Channel& channel)
 	{
-		m_invitationModeration ^= true;
-		client.createMessage(message.channelId(), m_invitationModeration ? "Invitations will be deleted!" : "Invitations won't be deleted!");
+			QStringList args = message.content().split(' ');
+			UmikoBot::VerifyAndRunAdminCmd(client, message, channel, 1, args, true, [this, &client, channel, message, args]()
+			{
+				m_invitationModeration ^= true;
+				client.createMessage(message.channelId(), m_invitationModeration ? "Invitations will be deleted!" : "Invitations won't be deleted!");
+			});
 	});
 }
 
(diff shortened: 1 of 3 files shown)


src/modules/ModerationModule.cpp
```diff
@@ -1,32 +1,55 @@
...
+			if (message.content().contains("https://discord.gg/", Qt::CaseInsensitive))
+			{
+				auto authorID = message.author().id();
+				::Permissions::ContainsPermission(client, channel.guildId(), message.author().id(), CommandPermission::ADMIN,
+					[this, message, &client, authorID](bool result)
+				{
+					if (!result)
+					{
+						client.deleteMessage(message.channelId(), message.id());
+						client.createMessage(message.channelId(), "**<@" + QString::number(authorID) + ">, invitation link of servers aren't allowed in any channels on this server. Please take it to DMs!**");
```
Gaztin: I would love it if instead the author would get a DM with the message they sent so that they have a chance to edit out the link. They might have spent some time writing the message and it would be a shame for the message to just be yeeted into the void for a small misunderstanding.

So if you send them two DMs. One explaining that the message was deleted because it contained an invite link, and another with the entire message in it 😊

Gaztin (changes requested):
Looks solid, just one small thing.

src/modules/ModerationModule.cpp
ridhwandll: Perfectly Done 😊

src/modules/ModerationModule.cpp
```diff
@@ -30,12 +30,18 @@ void ModerationModule::OnMessage(Client& client, const Message& message)
...
 					{
 						client.deleteMessage(message.channelId(), message.id());
-						client.createMessage(message.channelId(), "**<@" + QString::number(authorID) + ">, invitation link of servers aren't allowed in any channels on this server. Please take it to DMs!**");
+						UmikoBot::Instance().createDm(authorID)
+							.then([authorID, &client, message](const Channel& channel)
+								{
+									client.createMessage(channel.id(), "**<@" + QString::number(authorID) + ">, invitation link of servers aren't allowed in any channels on this server. Please take it to DMs!**");
+									client.createMessage(channel.id(), "**Here is your message which you posted in the server:**\n" + message.content());
+								});
+						client.createMessage(channel.id(), "**<@" + QString::number(authorID) + ">, invitation link of servers aren't allowed in any channels on this server. Please take it to DMs!**");
```
Gaztin: I don't think we need to send this message. They should already have gotten a DM about it, so the only reason for this public message would be for others to see it, which I see no reason for. It would clutter the advertise channel for example!

Gaztin (changes requested):

src/modules/ModerationModule.cpp
ridhwandll: What if their DM is off, from the server?

ridhwandll: Like this 
image

Gaztin: In my opinion, if they have opted out of getting DMs from server members, they have opted out of being informed about their participation in the server at a whole.
If someone behaves badly and I try and reach out to them in DMs to try and solve things, I don't wait around for them to accept a friend request. I take action immediately. Not saying anything is valid in this case.

Gaztin (approved):

PR #58, opened 2020-12-19
Improvements to the github command

masterneko:
## What's New?
- You can now use `--repository` to view a respository's details. 
eg. `!github TheChernoCommunity/UmikoBot` or `!github --repository TheChernoCommunity/UmikoBot`
- You can search for a repository too. eg. `!github --search linux`
- The language no longer shows blank on some repositories
- Added GitHub logo

### DIFF
--- a/src/core/Utility.h
+++ b/src/core/Utility.h
@@ -115,6 +115,7 @@ namespace utility
 			constexpr auto WE_SMART = "<:wesmart:777299810658680892>";
 			constexpr auto AANGER = "<:aanger:777298666485841921>";
 			constexpr auto SHOPPING_BAGS = ":shopping_bags:";
+			constexpr auto GITHUB = "<:github:777287502829846606>";
 		}
 	}
 
--- a/src/core/Utility.h
+++ b/src/core/Utility.h
@@ -82,6 +82,9 @@ namespace utility
 			}
 			
 			constexpr auto TICKETS = ":tickets:";
+			constexpr auto STAR = ":star:";
+			constexpr auto LAPTOP = ":computer:";
+			constexpr auto DESKTOP = ":desktop_computer:";
 
 			constexpr auto REGIONAL_INDICATOR_A = ":regional_indicator_a:";
 			constexpr auto REGIONAL_INDICATOR_B = ":regional_indicator_b:";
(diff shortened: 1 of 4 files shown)


src/core/Utility.h
```diff
@@ -82,6 +82,9 @@ namespace utility
 			}
 			
 			constexpr auto TICKETS = ":tickets:";
+			constexpr auto STAR = ":star:";
+			constexpr auto LAPTOP = ":computer:";
+			constexpr auto DESKTOP = ":desktop_computer:";
```
ridhwandll: Can you make the emoji name and the variable name same, to avoid confusion and make it easy to use? 😄 
```cpp
constexpr auto COMPUTER= ":computer:";
constexpr auto DESKTOP_COMPUTER = ":desktop_computer:";
```

ridhwandll (changes requested):

src/modules/FunModule.cpp
```diff
@@ -615,19 +667,67 @@ FunModule::FunModule(UmikoBot* client) : Module("funutil", true), m_memeChannel(
...
+
+			m_GithubManager.get(QNetworkRequest(QUrl("https://api.github.com/search/repositories?q=" + QString(QChar((char)ch(rng))))));
+		}
+		else if(m_GithubFlag == GITHUB_SEARCH)
+		{
+			m_GithubManager.get(QNetworkRequest(QUrl("https://api.github.com/search/repositories?q=" + repo)));
+		}
+		else if(m_GithubFlag == GITHUB_REPOSITORY)
+		{
+			m_GithubManager.get(QNetworkRequest(QUrl("https://api.github.com/repos/" + repo)));
```
Gaztin, line 729: This `get` function returns a `QNetworkReply*` that you can use to listen to its `finished` signal, instead of shoving all functionality in the QNetworkAccessManager's finished signal!

src/modules/FunModule.h
```diff
@@ -40,11 +40,27 @@ class FunModule : public Module, public QObject
...
+		QString language;
+		int stars;
+		QString url;
+	};
+
 	//! A list of roles (for each server) that have been given poll 
 	//! creation access
 	QMap<snowflake_t, QList<snowflake_t>> m_pollWhitelist;
 	QNetworkAccessManager m_MemeManager, m_GithubManager;
+	GithubFlag m_GithubFlag = GITHUB_RANDOM;
```
Gaztin, line 63: This probably shouldn't be be saved as a member variable. What happens if two !github requests are made at once, each with different flags?

Gaztin (changes requested):

PR #62, opened 2021-03-14
Small update in Moderation module

ridhwandll:
Now the bot deletes the `discord.gg` containing messages too

### DIFF
--- a/src/modules/ModerationModule.cpp
+++ b/src/modules/ModerationModule.cpp
@@ -26,7 +26,7 @@ void ModerationModule::OnMessage(Client& client, const Message& message)
 	{
 		if (m_invitationModeration)
 		{
-			if (message.content().contains("https://discord.gg/", Qt::CaseInsensitive))
+			if (message.content().contains("https://discord.gg/", Qt::CaseInsensitive) || message.content().contains("discord.gg", Qt::CaseInsensitive))
 			{
 				auto authorID = message.author().id();
 				::Permissions::ContainsPermission(client, channel.guildId(), message.author().id(), CommandPermission::ADMIN,


src/modules/ModerationModule.cpp
```diff
@@ -26,7 +26,7 @@ void ModerationModule::OnMessage(Client& client, const Message& message)
 	{
 		if (m_invitationModeration)
 		{
-			if (message.content().contains("https://discord.gg/", Qt::CaseInsensitive))
+			if (message.content().contains("https://discord.gg/", Qt::CaseInsensitive) || message.content().contains("discord.gg", Qt::CaseInsensitive))
```
Gaztin, line 29: First of all, the first term is redundant. if a message contains "https://discord.gg/" then it also contains "discord.gg".

second of all, just sending "discord.gg" as a message is going to get flagged, but it doesn't actually present an invitation widget. What we want to look for is:
`<zero or two or more non-whitespace characters>discord.gg/<one or more non-whitespace characters>`

Examples of messages that won't present an invitation widget:
* discord.gg/
* xdiscord.gg/test

Examples that will:
* discord.gg/x
* xydiscord.gg/test

Gaztin (changes requested):

src/modules/ModerationModule.cpp
ridhwandll: > First of all, the first term is redundant. if a message contains "https://discord.gg/" then it also contains "discord.gg".

 I thought the message have to be exactly same with the user input, that is "https://discord.gg/" and discord.gg are different, I see, I am wrong 🤔 

> second of all, just sending "discord.gg" as a message is going to get flagged, but it doesn't actually present an invitation ...

Okay I will try to fix that 👍

Gaztin: > I thought the message have to be exactly same with the user input, that is "https://discord.gg/" and discord.gg are different, I see, I am wrong 🤔

The reason it's redundant is we're using `contains` and not `==`

ridhwandll:
I am currently busy on another project. I don't want a dead pull request here, so I am closing this. I will open another one when I get enough time 😄

PR #63, opened 2021-06-01
Fixing a minor grammar mistake.

hannes-harnisch:
"More rich" is not the right comparative. "Richer" is. I hope it's not pedantic, it has just been bothering me for a while. 😶


ridhwandll (approved):
Nice catch, thanks!

FancyKillerPanda (approved):

PR #64, opened 2021-07-05
Moderation

FancyKillerPanda:
Added `!warn` and a way to see past warnings (`!warnings` or `!warnings-all`). Also changed some permissions stuff to allow people with moderator permissions to use these types of commands, rather than just admins.


ridhwandll:
Looks good to me! 🙂

ridhwandll (approved):

Gaztin:
LGTM, fix the build error on Linux and I will merge it!

PR #65, opened 2021-07-29
Adding check for bots in donate and steal commands

hannes-harnisch:
As discussed before. However I'm not entirely convinced it will build right away. I had some difficulties setting up Qt for compilation.


Gaztin:
LGTM

PR #66, opened 2021-08-06
Fixing bug with improperly printed emoji

hannes-harnisch:
The angry ping emoji didn't print correctly because it references an incorrect ID. This fixes that.

### DIFF
--- a/src/modules/CurrencyModule.cpp
+++ b/src/modules/CurrencyModule.cpp
@@ -873,7 +873,7 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
 			}
 			if (user.bot())
 			{
-				client.createMessage(message.channelId(), "Just _who_ do you think supplies your money here? Donate it to someone else! :angryping:");
+				client.createMessage(message.channelId(), QString("**Just _who_ do you think supplies your money here? Donate it to someone else!** <:") + utility::consts::emojis::reacts::ANGRY_PING + QString(">"));
 				return;
 			}
 		}


src/modules/CurrencyModule.cpp
```diff
@@ -873,7 +873,7 @@ CurrencyModule::CurrencyModule(UmikoBot* client) : Module("currency", true), m_c
 			}
 			if (user.bot())
 			{
-				client.createMessage(message.channelId(), "Just _who_ do you think supplies your money here? Donate it to someone else! :angryping:");
+				client.createMessage(message.channelId(), "**Just _who_ do you think supplies your money here? Donate it to someone else!** :angryping~1:");
```
Gaztin: You should use the pre-existing emoji ID in src/core/Utility.h instead. There's an ANGRY_PING constant.
Remember to wrap it in tags for message text, though:
```cpp
"<:" + ANGRY_PING + ">"
```

Gaztin (changes requested):

src/modules/CurrencyModule.cpp
hannes-harnisch: Ahh okay, I was under the impression this would work since other messages (in this same source file) also just use quotes like ``:police_officer:`` or ``:unlocked:``. And ``:angryping~1:`` seems to exist on the server.

Gaztin: It's kind of weird how the emoji system works, but the reason we can do `:police_officer:` is because it's a system emoji and not a custom one!

PR #68, opened 2021-09-29
Put modules into separate subdirectories

masterneko:
Some modules have become quite large.﻿

I have separated the modules into seperate files and subdirectories to simplify the development process.


FancyKillerPanda:
Looks alright to me, though i'm not sure if this is *too many* folders. We could maybe combine User, Timezone, Level, and Currency all to be under the User folder. We could also split the different events into separate files within the Event directory, but that can be done later.

masterneko:
I see there are conflicts. Do you want me to just close the PR or fix the conflicts?

